PR #66, opened 2016-08-31
Improve valuetype textures

Yulife:


rubensworks:
Awesome, the new textures look great!

Could you have another look at the number and/or string textures though?
They look very similar now, I already have a hard time distinguishing them, so this may be very problematic for a color blind person...

Yulife:
Done

rubensworks:
Any chance you could keep the yellow/orange color of the number texture, to stay similar to the integer, double and long?
It would be great if you just changed the shape of the string texture, possibly similar to what it was before.

Yulife:
How's this? :)

rubensworks:
Looks good! :-)
I'll have a look to see how it looks ingame.

rubensworks:
Looks great ingame as well, thanks!

I'll be releasing a new version today, this change will be included.

PR #94, opened 2016-12-19
Issue 93: Don't try to place cable when using an offset

Runesmacher:
Added a extra bboolean to check if the function is called with the offsett added to the position.


rubensworks:
Seems to work great, thanks!

PR #96, opened 2016-12-20
Issue 95: Fixed players being pushed off squeezer and basin

Runesmacher:


Runesmacher:
Not sure about all tehe commits or how this would be more correct logging wise.

PR #104, opened 2017-01-04
Improved item collecting from drying basin

### LINKED ISSUE #97: Right clicking on drying basin only possible with emty hand or same item
When getting items out of the drying basin, the item will only go in the current hand, even if there is already a item stack for this item in the inventory. You can also not click it with a non-empty hand

forge 2116
cyclopscore 0.8.2-432
integrateddynamics 0.6.5

Runesmacher:
Fixed #97

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/block/BlockDryingBasin.java
+++ b/src/main/java/org/cyclops/integrateddynamics/block/BlockDryingBasin.java
@@ -60,17 +60,22 @@ public class BlockDryingBasin extends ConfigurableBlockContainer implements IMac
         if(world.isRemote) {
             return true;
         } else {
-            ItemStack itemStack = player.inventory.getCurrentItem();
             TileDryingBasin tile = TileHelpers.getSafeTile(world, blockPos, TileDryingBasin.class);
             if (tile != null) {
+                ItemStack itemStack = player.inventory.getCurrentItem();
                 IFluidHandler itemFluidHandler = FluidUtil.getFluidHandler(itemStack);
                 SingleUseTank tank = tile.getTank();
                 ItemStack tileStack = tile.getStackInSlot(0);
-                if ((itemStack == null || (ItemStack.areItemsEqual(itemStack, tileStack) && ItemStack.areItemStackTagsEqual(itemStack, tileStack) && itemStack.stackSize < itemStack.getMaxStackSize())) && tileStack != null) {
+
+                if ((itemStack == null || (ItemStack.areItemStacksEqual(itemStack, tileStack) && tileStack.stackSize < tileStack.getMaxStackSize())) && tileStack != null) {
                     if(itemStack != null) {
                         tileStack.stackSize += itemStack.stackSize;
                     }
                     player.inventory.setInventorySlotContents(player.inventory.currentItem, tileStack);
+                    tile.setInventorySlotContents(0, null);
+                    tile.sendUpdate();
+                    return true;
+                } else if(player.inventory.addItemStackToInventory(tileStack)){
                     tile.setInventorySlotContents(0, null);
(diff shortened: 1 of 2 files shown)


rubensworks:
Haven't tested this myself yet, but does this still make sure that items will first be added to the currently held itemstack before the rest of the inventory?
If yes, then this change should probably also be done to the squeezer.

(And later on also the EvilCraft purifier...)

Runesmacher:
Updated so the hand is always prioritized. After that the first available stack will be filled.
Also updated the squeezer

src/main/java/org/cyclops/integrateddynamics/block/BlockDryingBasin.java
```diff
@@ -60,20 +60,25 @@ public boolean onBlockActivated(World world, BlockPos blockPos, IBlockState bloc
...
+
+                if ((itemStack == null || (ItemStack.areItemStacksEqual(itemStack, tileStack) && tileStack.stackSize < tileStack.getMaxStackSize())) && tileStack != null) {
                     if(itemStack != null) {
                         tileStack.stackSize += itemStack.stackSize;
                     }
                     player.inventory.setInventorySlotContents(player.inventory.currentItem, tileStack);
                     tile.setInventorySlotContents(0, null);
                     tile.sendUpdate();
                     return true;
+                } else if(player.inventory.addItemStackToInventory(tileStack)){
```
rubensworks, line 78: Why this separate if branch? It would seem to me that this can be merged with the above one, to avoid duplicate code.

Runesmacher: becous player.inventory.addItemStackToInventory(tileStack) is actually adding the item to the stack whereas the previous line is only checking the stack and adding it after the facrt, combining these could create a duplicate addition

rubensworks: I'll try to have a look at it later today.

src/main/java/org/cyclops/integrateddynamics/block/BlockDryingBasin.java
```diff
@@ -60,20 +60,25 @@ public boolean onBlockActivated(World world, BlockPos blockPos, IBlockState bloc
...
-            ItemStack itemStack = player.inventory.getCurrentItem();
             TileDryingBasin tile = TileHelpers.getSafeTile(world, blockPos, TileDryingBasin.class);
             if (tile != null) {
+                ItemStack itemStack = player.inventory.getCurrentItem();
                 IFluidHandler itemFluidHandler = FluidUtil.getFluidHandler(itemStack);
                 SingleUseTank tank = tile.getTank();
                 ItemStack tileStack = tile.getStackInSlot(0);
-                if ((itemStack == null || (ItemStack.areItemsEqual(itemStack, tileStack) && ItemStack.areItemStackTagsEqual(itemStack, tileStack) && itemStack.stackSize < itemStack.getMaxStackSize())) && tileStack != null) {
+
+                if ((itemStack == null || (ItemStack.areItemStacksEqual(itemStack, tileStack) && tileStack.stackSize < tileStack.getMaxStackSize())) && tileStack != null) {
```
rubensworks, line 70: This case never occurs, because of `ItemStack.areItemStacksEqual` also checking stacksizes.
It appears however that the second case you added with `addItemStackToInventory` is enough, and the vanilla code correctly handles this as intended.

So you can remove this first if-block completely. You can double check to make sure I didn't forget any edge-case.
That if-block can then also be removed for the Squeezer.

rubensworks (changes requested):

Runesmacher:
Closed in favor of #110

PR #236, opened 2017-08-08
Make enriched menril logs silk touchable

josephcsible:
It's rather odd to have a block that can be infinitely obtained in block form in survival, and obtained as an item from the creative inventory, but not obtainable in any way as an item in survival. Let's fix that oddity.


rubensworks:
Thanks!

PR #266, opened 2017-09-03
Fix VariableFacadeHolderConfig using the wrong implementation class

josephcsible:


rubensworks:
Oops, that's wrong indeed, thanks! :-)

PR #267, opened 2017-09-04
Fix some type-safety issues

josephcsible:


rubensworks:
Cool, thanks!

@josephcsible Any reason why you're doing all of this cleanup?
Are you implementing something on top of this?
Or do you just like cleaning up code? ;-)

josephcsible:
After a crash due to [heap pollution](https://en.wikipedia.org/wiki/Heap_pollution) caused me a lot of headaches, I'm trying to minimize the amount of raw type and unchecked cast warnings in all of the mods I use, so there's nowhere for such bugs to hide.

PR #320, opened 2017-10-23
Fix #319: override the non-deprecated version of getDrops

### LINKED ISSUE #319: [1.12] Industrial Foregoing's Plant Harvester doesn't get berries or crystalised chunks from Menril trees
I assumed the problem was with Industrial Foregoing but it's author points out the problem may be on this side in Buuz135/Industrial-Foregoing#74.

josephcsible:


josephcsible:
Also needs CyclopsMC/CyclopsCore#81

PR #332, opened 2017-10-29
Add channels

josephcsible:


src/main/java/org/cyclops/integrateddynamics/api/network/IPositionedAddonsNetwork.java
```diff
@@ -51,11 +51,12 @@
 
     public static class PrioritizedPartPos implements Comparable<PrioritizedPartPos> {
         private final PartPos partPos;
-        private final int priority;
+        private final int priority, channel;
```
rubensworks, line 54: The `compareTo` should also be extended to include this channel field, probably after the positions are compared. Even though it is not needed at the moment, in the future it may be possible to have multiple parts on the same positions, but on different channels.

src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
```diff
@@ -394,6 +398,15 @@ public ITemperature getOutput(Pair<PartTarget, IAspectProperties> input) {
 
     public static final class Network {
 
+        public static final IAspectPropertyTypeInstance<ValueTypeInteger, ValueTypeInteger.ValueInteger> PROPERTY_CHANNEL =
```
rubensworks, line 315: Not sure if this property is needed. I think it should only be possible for aspects to interact with things that are on the same channel as the channel its part is in.

It's indeed more flexible like this, but perhaps it's a bit too flexible. Thoughts?

src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
```diff
@@ -394,6 +398,15 @@ public ITemperature getOutput(Pair<PartTarget, IAspectProperties> input) {
...
+                new AspectPropertyTypeInstance<>(ValueTypes.INTEGER, "aspect.aspecttypes.integrateddynamics.integer.channel.name");
+        public static final IAspectProperties PROPERTIES = new AspectProperties(ImmutableList.<IAspectPropertyTypeInstance>of(
+                PROPERTY_CHANNEL
+        ));
+        static {
+            PROPERTIES.setValue(PROPERTY_CHANNEL, ValueTypeInteger.ValueInteger.of(IChanneledNetwork.WILDCARD_CHANNEL));
```
rubensworks, line 321: If we keep this property, it might be better to set the default channel to channel 0. Because if players want to use a different channel instead, _all_ channels will have to be updated from wildcard to something else, instead of only changing one channel 0 to something else.

src/main/java/org/cyclops/integrateddynamics/network/EnergyBatteryNetworkElement.java
```diff
@@ -62,7 +63,7 @@ public void addDrops(List<ItemStack> itemStacks, boolean dropMainElement) {
 
     @Override
     public boolean onNetworkAddition(INetwork network) {
-        return NetworkHelpers.getEnergyNetwork(network).addPosition(PartPos.of(getPos(), null), 0);
+        return NetworkHelpers.getEnergyNetwork(network).addPosition(PartPos.of(getPos(), null), 0, 0);
```
rubensworks: For consistency, use `IChanneledNetwork.DEFAULT_CHANNEL` here as well.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetworkElement.java
```diff
@@ -64,6 +64,11 @@ public int getPriority() {
     }
 
     @Override
+    public int getChannel() {
+        return hasPartState() ? part.getChannel(getPartState()) : 0;
```
rubensworks: For consistency, use `IChanneledNetwork.DEFAULT_CHANNEL` here as well.

src/main/java/org/cyclops/integrateddynamics/core/network/EnergyChannel.java
```diff
@@ -0,0 +1,90 @@
+package org.cyclops.integrateddynamics.core.network;
+
+import org.cyclops.integrateddynamics.GeneralConfig;
+import org.cyclops.integrateddynamics.api.network.IChanneledNetwork;
+import org.cyclops.integrateddynamics.api.network.IPositionedAddonsNetwork.PrioritizedPartPos;
+
+import net.minecraftforge.energy.IEnergyStorage;
+
+public class EnergyChannel implements IEnergyStorage {
+    final EnergyNetwork network;
```
rubensworks: I tend to make these things private by default, any reason why you are not doing that? (Just a question, you shouldn't change this if you have a good reason)

src/main/java/org/cyclops/integrateddynamics/core/network/EnergyChannel.java
```diff
@@ -0,0 +1,90 @@
...
+    final int channel;
+
+    EnergyChannel(EnergyNetwork network, int channel) {
```
rubensworks: Similar question here, any reason why this constructor is package-private and not just public?

rubensworks (changes requested):
Looks good! Really nice and clean implementation.

Just a few comments, feel free to comment if you don't agree with anything.

src/main/java/org/cyclops/integrateddynamics/core/network/EnergyChannel.java
josephcsible: Nothing but EnergyNetwork::getChannel should ever call it.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetworkElement.java
josephcsible: Good catch.

src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
josephcsible: The rest of the properties here are about the entire network, though, so shouldn't this one also be that way by default?

src/main/java/org/cyclops/integrateddynamics/core/network/EnergyChannel.java
rubensworks: That's true, but I'm just thinking in terms of extensibility. If some external mod would like to create its own energy channel for whatever reason, should it be able to use this implementation? I think the answer is yes, unless something could go wrong that way?

src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
rubensworks: Yes, makes sense. I thought this property had another purpose, you can ignore my comment. (and also the one in IT then)

rubensworks (approved):

src/main/java/org/cyclops/integrateddynamics/capability/network/NetworkCapabilityConstructors.java
```diff
@@ -21,12 +23,13 @@ public void onNetworkLoad(AttachCapabilitiesEventNetwork event) {
         INetwork network = event.getNetwork();
         PartNetwork partNetwork = new PartNetwork();
         EnergyNetwork energyNetwork = new EnergyNetwork();
+        IEnergyStorage energyChannel = energyNetwork.getChannel(IChanneledNetwork.DEFAULT_CHANNEL);
         event.addCapability(new ResourceLocation(Reference.MOD_ID, "partNetwork"),
                 new DefaultCapabilityProvider<>(() -> PartNetworkConfig.CAPABILITY, partNetwork));
         event.addCapability(new ResourceLocation(Reference.MOD_ID, "energyNetwork"),
                 new DefaultCapabilityProvider<>(() -> EnergyNetworkConfig.CAPABILITY, energyNetwork));
         event.addCapability(new ResourceLocation(Reference.MOD_ID, "energyStorageNetwork"),
```
josephcsible, line 31: As I look at this again, what do we even need this capability for? Can we just remove it?

rubensworks: It is indeed not used anywhere ATM, but it makes sense to have this here IMO. As the network can handle items using the standard IH capability. Maybe some other mod in the future may have a use for this. (It doesn't cause any issues to have this here in any case)

josephcsible: But wouldn't you need to use our API to even get the network object at all, before you could get its capabilities? And if you are using our API, shouldn't you pass a channel rather than getting the default one implicitly?

rubensworks: It can be convenient for mod developers, so I won't be changing it.

PR #337, opened 2017-10-31
Add rftools and rftoolscontrol constants

josephcsible:


rubensworks:
Why is this needed here? Can't this be added in compat instead?

josephcsible:
Funny, I was going to ask you the same thing. I assumed you wanted them here since the rest of compat's constants seem to live here. If not, just close this and I'll add them there with the actual change.

rubensworks:
Yes, that was an oversight on my part when porting all modcompats to IDCompat. At some point in the future this should be cleaned up, but there are plenty of more urgent things at the moment :-)

PR #344, opened 2017-11-05
Simplify the redstone clock so that duty cycles > 50% work

### LINKED ISSUE #318: [1.12.2]Redstone clock has a bug. Specal pulse setting will not outout pulse.
You can try to modify the pulse interval & pulse length like this. (3,2),(7,5),(10,5),(12,6)...
If pulse interval is less than or equal to two times of pulse length, will always out put true.

josephcsible:
Fixes #318 

Also, revert the commits that added the text that this doesn't work.


rubensworks:
Sure, why not, and this makes the aspect implementation a lot simpler as well :-)

PR #355, opened 2017-11-09
Add string operators, localisation, tests.

tekacs:
> Before submitting a pull request containing a new feature, please discuss this first with one of the lead developers.

Apologies for not contacting you before writing this - I understand if you have different plans and that this branch isn't useful/already done/wrong.

I noticed that when string operators were mentioned on Reddit and in issues here, they were 'probably' in the next release, so I wanted to campaign for making it 'definitely'. :smile:

This branch adds the operators:

* `contains`: If the string contains the given substring.
* `contains_regex`: If the string contains a substring matching the given regular expression
* `starts_with`: If the string starts with the given substring.
* `ends_with`: If the string ends with the given substring.
* `split_on`: Get a list containing pieces of the given string, split on the given (literal) delimiter.
* `split_on_regex`: Get a list containing pieces of the given string, split on the given regular expression.
* `substring`: Get a substring of the given string, from the given index (inclusive) to the given index (exclusive).
* `regex_group`: Match the given string against the given regular expression and return the group at the given index.

I've written (positive and negative) tests for all of these (which caught a thing or two) and I believe these are all exception-safe. I've tried to closely match the style of the surrounding code.

The thinking in providing regexes is that they prevent a massive proliferation of additional functions and no-one should find themselves wanting and come requesting more operators (if something isn't provided, the three `_regex` functions should mostly have you covered).

I'm not sure if you have a 'maximum flexibility/power/balance' concern, but since regexes can be poorly implemented (more expensively) using the other commands I hoped they would be okay. To the best of my knowledge, other than DoS (which I believe you could pull off with other operators), the Java regex engine is safe to expose in the way I have done here (and so should be server-safe).

I would have hoped to return `null` from `regex_group` when the regex as a whole failed to match, to be tested with the `is_null` operator, but I couldn't see any instances of `null` being used in this way and so I made it an error.

I skipped `indexOf` due to lacking a use for it, but it should be easy to add following the above pattern.

Errors aren't passed through localisation right now - I can move the error messages to localisation if that's appropriate in this context.

Again, apologies for not asking in advance (I got somewhat carried away) - hopefully this matches some approximation of what you had planned to provide!


rubensworks:
Oh, cool! I'll have a look at it in detail later today.

rubensworks (approved):
Looks great, thank a lot for this!

Will merge this once I prepare the next feature release (0.10.0).

rubensworks:
Next to your operators, I also intended to implement support for indexOf and match (just a string+regex to boolean).
Also, your implementation of regex_group is definitely useful, but I'd also implement another variant that returns a string list of all groups, which is just slightly more flexible.

Feel free to have a look at these yourself, otherwise I will gladly implement them myself.

tekacs:
@rubensworks Thanks, sounds good! Replying to your comments,

* I believe `contains_regex` satisfies `match`? I originally implemented `contains_regex` as `matches_regex`.
  I switched it to `contains` because it's trivial to convert the former to the latter by adding `^` and `$` or `\A` and `\z` to the beginning and end of your regex (so `^hello` instead of `hello`). If you'd still like `matches_regex` back, it's just `contains_regex` with `s/find/matches/`.
* I've added `index_of`. I also added `index_of_regex` for situations like `map(["Alchemical Iron Ingot", "Resonating Cobalt Ingot"], indexOfRegex(_, "\S+ Ingot"))`.
* I added `replace` and `replace_regex` - `.replaceAll` from Java - I left out single-element replace because you can trivially build that yourself with `regex_groups` and `concat`. Replacing all instances is much harder.
* I added `regex_groups` as you requested - that makes much more sense than my `regex_group` implementation, since it handles the empty case much better - thanks! I've left `regex_group` in since it's still a very handy shortcut for simple text extraction.

As previously, please feel free to trim/adjust these according to your taste. :smile:

tekacs:
The operators seem to work fine at runtime and enable some really fun new functionality. :smile:

I'm looking forward to the release, so I can start solving Minecraft problems in https://xkcd.com/208/ style.

One small question/caveat: having tried to use the `contains_regex` function as a predicate [1], I noticed that the first thing I had to do was `flip` it, in order to hard-code the `regex` input whilst leaving the `string` (to match) input free.

I used this argument order in order to re-use the existing UI, which puts one slot on the left and two to the right in `INFIX_2`. You might want to have a look over and determine whether the parameter orders herein are the right ones, since you've used this in game rather more than me (I don't mind flip in Haskell, but in game it's more unwieldy and costly).

[1]: Use case:

(1) Operator: String Contains Regex (String -> String) -> Boolean
(2) Flip: applied to (1) to flip the two string parameters
(3) Apply: applied to (2) to partially apply it with the regex I want to search for (`(Alchemical|Superior)`)
(4) Operator: name (Item -> String)
(5) Pipe: function compose (4) and (1)

... then put (5) into the 'Import Items Predicate' slot of the item importer. This is (the applies are implicit here):

`predicate = flip containsRegex "(Alchemical|Superior)" . name`

And so on even larger, 

`extract = flip (flip regexGroup "check(thing)" . name) 1`
`predicate = ("thing" ==) . extract`

rubensworks:
Awesome, thanks!

> I believe contains_regex satisfies match?

Yes, it can be done indeed. But it might be convenient for non-expert players not having to add `^` and `$`. So I'd add a match operator in any case.

>One small question/caveat: having tried to use the contains_regex function as a predicate [1], I noticed that the first thing I had to do was flip it, ...

It definitely makes sense to change the input order of this operator (and possible the others as well). Because for partially applied operators, the regex will in most cases be the first one that will be applied for further reusable applications. So feel free to change this :-)

josephcsible:
I'd like to take a good look at this to make sure that it's not vulnerable to ReDOS or catastrophic backtracking. See http://blog.mgm-tp.com/2012/06/regexp-java-puzzler-2/ for an example of what this is and why it's so dangerous. See also https://www.regular-expressions.info/redos.html (specifically "Handling Regexes Provided by The User").

josephcsible (changes requested):
Yeah, this is really vulnerable. Using contains_regex on `^(aa|aab?)*$` and `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` (63 a's, one less than the max length of 64) will hang the server thread. I recommend that you split the regex operators into a separate PR so that everything else can be merged while we figure out a solution for them.

josephcsible:
Side note, if you're going to flip the arguments of "contains", I think "matches" or "is in" would be a better name.

rubensworks:
@josephcsible Thanks for investigating this!

I suggest creating something like an [`InterruptibleRegex`](http://planet.jboss.org/post/how_to_interrupt_a_long_running_infinite_java_regular_expression), with a certain (configurable) timeout value.

@tekacs I don't think it's required to split of the regex operators to a different PR though, it'll still take a while before I'll prepare the next feature release (hopefully not too long though).

tekacs:
@josephcsible Thanks for investigating this. As I mentioned in my original post, DoS is the only strategy that I was aware of that could cause problems here, though I dismissed it upon assuming that other operators within Integrated Dynamics could also have a DoS effect.

That not in fact being the case, something like @rubensworks' suggested `InterruptibleRegex` (something wtih a timeout value) is what I had in mind as a fallback (though I was thinking of this from a Scala perspective, where the solution is rather less messy to put together :smile:).

I'd like to ask though - given my original question about other operators too - would it perhaps be more useful to simply provide the option to run /some/ operators on a separate, interruptible thread (with their co-operation)? I ask only because I wonder if this is bound to come up again later, in another operator or other user-driven action.

Whatever the case, I'll try to make time to get to addressing this in the coming week (though no promises, just to be sure).

josephcsible:
I actually think I'd prefer something like [Lua patterns](https://www.lua.org/manual/5.3/manual.html#6.4.1) or something else that isn't powerful enough to cause the problem than just a regex with a time limit. Re other operators, I'd rather make sure that they're all fast than start dealing with threads.

rubensworks:
@tekacs Instead of adding a timeout safeguard to the regex operators, it might indeed be useful to just adds this safeguard to _all_ operators. So either this check can be added to the whole [operator chain](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.12/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/OperatorBase.java#L136), or to single [operator executions](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.12/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/IterativeFunction.java#L28). The latter (which I would prefer) could be done by creating a threaded `IOperatorValuePropagator` that adds a timeout safeguard, which would probably be better. This custom value propagator could only be used for operators that need them, which would avoid some synchronization overhead on operators that do not need them.

But as @josephcsible has said, perhaps we should first look into other, less expressive, regex languages to avoid this issue altogether.

@josephcsible Do you by any chance know of a Java library that implements Lua patterns (or something similarly safe)? We _could_ shadow in such a jar if the license allows it.

josephcsible:
> Do you by any chance know of a Java library that implements Lua patterns (or something similarly safe)?

Not offhand, but I can look into it.

tekacs:
> Instead of adding a timeout safeguard to the regex operators, it might indeed be useful to just adds this safeguard to all operators.

Yup definitely, that's what I was suggesting above.

> But as @josephcsible has said, perhaps we should first look into other, less expressive, regex languages to avoid this issue altogether.

I don't think Lua patterns (specifically) are the solution, for sure.

If this approach were to be taken, I would suggest ignoring other pattern languages and picking up a linear-time regular language matching implementation along the lines of https://github.com/google/re2j that ditches the backtracking implementation and limits to that which can be handled by an NFA (re2 is the well known solution in this space and is specifically used in cases like this - again as above I assumed wrongly when in a hurry, that the relative lack of dependencies on this package was for an important reason).

If nothing else, folks are much more likely to be familiar with (and not have to look up a manual for, and find online help and online tools to help them test) regexes than with other pattern languages (I say this as someone who would prefer a textbox and a PEG parser generator - but that would supremely user-unfriendly 😄).

josephcsible:
I didn't know of RE2/J until now, but now that I see it, I do like it better than Lua patterns.

tekacs:
This is the bit where I sneak in a PEG parser generator. 😈

K, I'll take a look at moving this to RE2/J instead.

I assume there's no packaging issues with simply adding it as a dependency?

rubensworks:
RE2/J indeed looks interesting, we can indeed look into using that.

> I assume there's no packaging issues with simply adding it as a dependency?

The license should allow us to shadow it in, as long as we retain the copyright notice. But this should be automatically be taken care of when shadowing.

rubensworks:
@tekacs FYI, I'll start preparing a feature release in a few weeks. Do you have time to look into changing your PR to use RE2/J? If not, either me or @josephcsible can look into it, or we can put this on hold until the feature release after the next one.

tekacs:
This should be a very quick change - I'll see if I can find the time this weekend.

rubensworks:
@tekacs Any news? I *might* do a release somewhere next week.

rubensworks:
@josephcsible Are you interested in taking one this up?

josephcsible:
I'll take a look at it.

josephcsible:
Closing as abandoned in favor of #447

PR #361, opened 2017-11-12
Use lambdas to vastly simplify AspectReadBuilders and AspectWriteBuilders

josephcsible:


rubensworks:
Awesome, thanks! This (together with lamba-izing the operators and aspects themselves) has been on my todolist for a while :-)

PR #370, opened 2017-11-14
Disable Charset - it's down

josephcsible:


rubensworks:
Are the Charset builds not available from somewhere else? If not, we could host it ourselves on bintray (like the RF API for 1.11: http://dl.bintray.com/cyclopsmc/thirdparty/)

I'd rather not remove Charset support for 1.10, as people might still be using Charset with ID on 1.10.

josephcsible:
I have no idea. Why don't you just self-host it?

rubensworks:
Closing this, since we don't maintain 1.10 anymore.

PR #372, opened 2017-11-14
Make equals input type checks less strict, Closes #365

### LINKED ISSUE #365: comparison's type check fails if one parameter is a variable of type any
#### Issue type:

- :bug: Bug

____
#### Short description:

When trying to create a comparison such as `!=` in the logic programmer, I get a type error and can't save the result if one parameter is a variable of type `any` (for instance, the result of a `?` choice) and the other has a specific type (say an integer constant)

#### Expected behaviour:

The type check always passes if one or both parameters are of type `any`, and I can put the result in a variable.

#### Actual behaviour:

A red X is shown next to the result slot, and if I put in a variable card, nothing is saved to it.
type error

1. Create a `?` choice variable where all the possible values it can take are integers
2. Create an integer constant variable
3. Try to create a `!=` comparison of the `?` choice and the constant

____
#### Versions:

- This mod: 1.12.2-0.9.5
- CyclopsCore: 1.12.2-0.10.22
- Minecraft: 1.12.2
- Forge: 14.23.0.2547

josephcsible:


rubensworks:
Thanks, but 1.11 is not needed. Barely anyone still plays 1.11. People have either moved on to 1.12, or they are still using 1.10.

Issue #317, opened 2017-10-21
[Suggestion]Allow to reset Redstone Clock


Flithor:
![image](https://user-images.githubusercontent.com/23412916/31850391-97022cbc-b683-11e7-88fe-bb3b3a890632.png)
Hope to can reset the start time of pulse in some way.

rubensworks:
Can't think of a good and easy way at the moment to make this configurable.

Flithor:
Add a setting “restart when Redstone high(true/false)”

rubensworks:
That's not possible, because this is a reader aspect, not a writer aspect.

Flithor:
In addition, redstone clock has a bug.
You can try to modify the pulse interval & pulse length like this. (3,2),(7,5),(10,5),(12,6)...
You will find it will no longer be a pulse. always output true.

Flithor:
I said is the reader settings, added in the pulse settings.

rubensworks:
> In addition, redstone clock has a bug.

Please make a separate issue for this.

Flithor:
> That's not possible, because this is a reader aspect, not a writer aspec

Or to accept the redstone strength is greater than the set value to reset the redstone clock counter.

Issue #197, opened 2017-06-17
Feature Request: Aspect Variable Cards


ricochet1k:
In my attempts to make an auto-crafting system with this mod, I have found myself wanting to modify the parameters of Aspects with variable cards.

For example: Use a counter of some sort to iterate over the slots of an Item Exporter's Export Item aspect. This could be used to, for example, export specific items to specific slots of an inventory, such as an Analog Crafter, from Extra Utilities 2 IIRC.

Other things this could be used for: changing the pulse delay of the redstone clock aspect, or changing the stack size of item export/import.

rubensworks:
This might be possible to implement, but I'm wondering if it's really required.
Could additional, more complex aspects, for certain parts not be sufficient?

ricochet1k:
Well certainly, you could add an aspect that only exports items if the target inventory doesn't have that many already. Or just add an autocrafting system. But the whole reason I'm using this mod is that it seems to just give me the bare minimum of functionality and then lets us piece together what is necessary. Coming up with a complicated system to walk over each slot of the inventory, inserting an item if it exists, pushing that item onto a stack to construct if it doesn't exist, that sounds tedious, but it's actually fun.

If I wanted an easy "magical" way of autocrafting things, I'd be using Refined Storage.

rubensworks:
That's not what I'm talking about. I am thinking about an additional aspect that is more expressive than for example the operator aspect from item exporters, which could for example also take a 'slot' parameter next to the 'item' parameter in the existing aspect.

ricochet1k:
Hmm, so something like returning a `(Int, Item)` tuple of slot and item? Or are you referring to an `Int -> Item` Operator? Or are you referring to an Aspect with multiple variable cards, one for item, one for slot, one for stack size, etc.?

If it's the third one, then I'm having a hard time seeing how that's different from allowing parameters to be configured with variables. Why not always do it that way? You can create constant variables pretty easily.

rubensworks:
I'm referring to the second one, so an `(Int -> Item) -> Boolean` operator.

ricochet1k:
That one makes me nervous. There are some very large inventories in some mods, such as Refined Storage or Colossal Chests. You could easily crash things or cause horrible lag if you try to use it on one of those.

[...]

PR #380, opened 2017-11-17
Add an offset time to the redstone clock

josephcsible:
This lets you have, for example, 4 redstone clocks, each one on for a different
1/4 of each second. Without this, they'd all be on for the same 1/4 of each
second. This partially solves #317; it will fully solve it once #197 is solved.


rubensworks (approved):
Awesome, really useful!

PR #388, opened 2017-11-20
Fix #331: don't crash over an aspect model not being found

### LINKED ISSUE #331: NPE if a model .json is missing
If a model .json is missing for an aspect (such as if you delete https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.12/src/main/resources/assets/integrateddynamics/models/aspect/read/integer/network/energy/max.json), Minecraft will crash with a NullPointerException when you try to look at it. We should handle this more gracefully, just showing the magenta-and-black "missing" checkerboard or something.

josephcsible:
This gives you a nice big error from Forge in the log, but lets you use the
aspect anyway. This will make it much easier to code new aspects and worry
about the models later.


rubensworks:
Nice, thanks!

PR #400, opened 2017-11-23
Use the right name when giving type errors from NBT

josephcsible:


rubensworks:
Thanks! Too much copy-pasting on my end...

PR #406, opened 2017-11-24
Fix #397: stop misusing hash codes in uniq and uniq_p

### LINKED ISSUE #397: uniq and uniq_p don't work as intended
https://github.com/CyclopsMC/IntegratedDynamics/blob/f025d16240d2b715abe7a1ab4084f6b5bae697f9/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java#L748-L756
https://github.com/CyclopsMC/IntegratedDynamics/blob/f025d16240d2b715abe7a1ab4084f6b5bae697f9/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java#L772-L774

~~Two~~Three problems are apparent with the above code:
1. If two objects have the same hash code, they'll be treated as equal whether or not they actually are.
2. It's possible that two objects that the predicate considers equal will have different hash codes. This means that even discounting problem 1, this isn't a total order, so the set may misbehave.
3. The order of elements isn't preserved.

josephcsible:
Some things to note:
1. uniq_p is now O(n^2), but as far as I can tell, there's no correct way to
   implement uniq_p on an unsorted list with only an equality predicate
   otherwise. We can work around this by either weakening uniq_p to only
   remove adjacent duplicate elements (like the uniq command on Linux), or we
   can make it require a comparison function rather than an equality function.
   (The latter would correspond nicely with a future sort function, should we
   ever want to implement that.)
2. uniq's implementation depends on us implementing hashCode() correctly on
   all types that implement IValue. It looks like we don't always do this.
   I'll fix this in a separate PR.

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -3,6 +3,7 @@ package org.cyclops.integrateddynamics.core.evaluate.operator;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 
 import lombok.Lombok;
 import net.minecraft.block.Block;
@@ -71,11 +72,12 @@ import org.cyclops.integrateddynamics.core.helper.Helpers;
 import org.cyclops.integrateddynamics.core.helper.L10NValues;
 import org.cyclops.integrateddynamics.core.helper.obfuscation.ObfuscationHelpers;
 
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.TreeSet;
 
 /**
  * Collection of available operators.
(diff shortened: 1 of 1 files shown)


rubensworks:
Assuming a correct implementation of hashcodes, I think adding all elements to a LinkedHashSet would be more efficient. Insertion complexity is `O(1)`. In practise this will be a bit more than adding to a simple list, but it will definitely be less than `O(n^2)`, as this can exploit the hashcode.

josephcsible:
For uniq that's true (and it's what I did), but for uniq_p it's not.

Specifically, it's not true because we have potentially "equal" elements that will have different hash codes, so we can't use any Hash* containers to hold anything.

rubensworks:
Indeed, my bad, was checking the wrong part of the code...

rubensworks:
I'll leave this open for now. There must be a way to get uniq_p below `O(n^2)` as well. Will think about it a bit.

josephcsible:
I'm actually fairly sure that it's impossible, and in any case, isn't slightly slow better than wrong?

josephcsible:
Clarification: I only believe it to be impossible if you don't want to change uniq_p's semantics. If you make either of the changes in my first comment, it's easy.

rubensworks:
Indeed, the black box-nature of the predicate makes it quite impossible to improve this. That's also the reason why IT can have performance issues when predicate-based operations are used for item transfer in large inventories.

PR #407, opened 2017-11-24
Clean up equals and add hashCode to Network

josephcsible:


src/main/java/org/cyclops/integrateddynamics/core/network/Network.java
```diff
@@ -178,7 +176,12 @@ public void initialize() {
 
     @Override
     public boolean equals(Object object) {
-        return object instanceof Network && areNetworksEqual(this, (Network) object);
+        return object instanceof Network && elements.equals(((Network)object).elements);
```
rubensworks, line 179: At some point (possibly 1.13), this should accept any kind of `INetwork`. (This method was written before `INetwork` was added)

josephcsible: Doesn't it already work right with them? If you're comparing a Network to an INetwork that's not a Network, how could they possibly be equal? An non-Network INetwork might not even have "elements".

rubensworks: Indeed, but not the other way around.

It would then have to be compared using `getElements()`, instead of the direct field.

PR #408, opened 2017-11-24
Clean up equals and add hashCode for ValueTypeListProxyBase and Value…

josephcsible:
…TypeOperator

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyBase.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyBase.java
@@ -82,10 +82,13 @@ public abstract class ValueTypeListProxyBase<T extends IValueType<V>, V extends 
 
     @Override
     public int hashCode() {
+        if(this.isInfinite()) return System.identityHashCode(this);
         int result = 1;
         result = 37 * result + getName().hashCode();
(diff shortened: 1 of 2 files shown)


src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyBase.java
```diff
@@ -77,6 +80,15 @@ public boolean equals(Object obj) {
         return Arrays.equals(o, o2);
     }
 
+    @Override
+    public int hashCode() {
+        int result = 1;
+        result = 37 * result + getName().hashCode();
+        result = 37 * result + getValueType().hashCode();
+        result = 37 * result + (!this.isInfinite() ? Arrays.hashCode(Iterables.toArray(this, Object.class)) : 0);
```
rubensworks: I think `Arrays.hashCode(Iterables.toArray(...))` can be made faster by manually calculating the hashcode using a simple foreach loop. That way we avoid allocating a new array.

rubensworks (changes requested):

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyBase.java
josephcsible: Done.

PR #409, opened 2017-11-24
Add more tests to uniq and uniq_p

josephcsible:
While working on #397, I discovered that they also have the problem of not
preserving order. Add tests for that problem as well as the collision problem.
Note that these tests will fail until #406 is merged.


josephcsible:
Now that #406 is merged, this can be merged without breaking the build.

PR #419, opened 2017-11-26
Make cable connections side aware

davenonymous:
The mod Compact Machines adds a block that allows each of its faces
to be "connected" to a different block in a different dimension.
Because each face could be a different network/grid/system entirely
it requires the capabilities to be sided to be "proxied".

With this commit the path checks ask for the proper side of the
neighboring block instead of passing null to the capability checker
/getter.

This should not have any bad effects on the given cable system and
allows Integrated Dynamics to work with Compact Machines and other
Forge Capability proxying blocks (like Phantom Faces?) that require
a sided capability.

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementCable.java
+++ b/src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementCable.java
@@ -26,7 +26,7 @@ public abstract class PathElementCable extends PathElementDefault {
         for (EnumFacing side : EnumFacing.VALUES) {
             if (getCable().isConnected(side)) {
                 BlockPos posOffset = pos.offset(side);
-                IPathElement pathElement = TileHelpers.getCapability(getPosition().getWorld(), posOffset, PathElementConfig.CAPABILITY);
+                IPathElement pathElement = TileHelpers.getCapability(getPosition().getWorld(), posOffset, side.getOpposite(), PathElementConfig.CAPABILITY);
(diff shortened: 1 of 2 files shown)


rubensworks:
Thanks!

Unfortunately, I don't think your changes will completely solve the problem. There are still several other calls to the path element and cable capabilities that do not pass the side. Furthermore, I'm afraid more invasive changes are required from our end.

For instance:
* `NetworkHelpers#initNetwork` passes a null side to retrieve the path element cap. This means that reinitializing the network would probably fail when connected with CM.
* `Cluster#fromNBT` similar as above.
* We'd probably want to remove/deprecate the non-sided `CableHelpers#getCable` method, and refactor all calls to it to use the sided-variant.

The `CableHelpers#getCable` changes should be easy. The path element cap changes will probably require significant changes as to how networks are stored, so we'll have to come up with a backwards-compatible solution for this.

davenonymous:
Uhh, I see. I was afraid that more work than this would be involved. It seems like it would require me to dive a lot deeper into your code base. And those would probably be changes you'd want to make yourself.

For Compact Machines I might be able to go a different route and add support for non-sided capabilities, but that would require 1:n or even n:m connections (I wouldn't know which outside block face to connect to), which is a whole other can of worms.

rubensworks:
Indeed, I'll better look into these changes myself. Thanks in any case for the effort. I'll try to look into resolving this via #420 for ID 0.10.x or 0.11.x.

PR #424, opened 2017-11-28
Add pack.mcmeta

josephcsible:


rubensworks:
This is not really needed here IMO. The fallback version (2) is still compatible for us in this case (if this is not the case, please let me know). In 1.13, the pack version is likely to update anyways, so let's change it then.

josephcsible:
Are you sure? From what I've read, the fallback version implies "US" rather than "us".

rubensworks:
Indeed, looks like I misread. I thought the legacy loader _ignored casing_, but instead, it requires _mixed casing_. It simply works now because Mac ignores casing.

Issue #429, opened 2017-12-01
Group ID of omni-directional connector not synced to client


josephcsible:
#### Issue type:

_Only keep the issue types that are applicable._

- :bug: Bug

____
#### Short description:
When crafting an omni-directional connector, its group ID isn't immediately synced to the client.

#### Expected behaviour:
A newly-crafted omni-directional connector should immediately have its group ID visible.

#### Actual behaviour:
A newly-crafted omni-directional connector does not have its group ID visible until you force the server to tell the client about the item again, such as by moving it around.

#### Steps to reproduce the problem:

1. Craft (or reset) an omni-directional connector, and place it in your inventory
2. Mouse over it, and note the lack of a group ID
3. Move it to a different slot in your inventory
4. Note that its group ID is now visible

____
#### Versions:

Git revision c5ed8ac, with all of the dependencies pulled by gradle for it

#### Log file:

Nothing relevant

josephcsible:
https://github.com/CyclopsMC/IntegratedDynamics/blob/edef106c1bfc26d0dc775144ffefee150aff87a0/src/main/java/org/cyclops/integrateddynamics/part/PartTypeConnectorOmniDirectional.java#L171

This is the cause. Is there a reason for only doing this on the server?

rubensworks:
@josephcsible Because only the server knows of these IDs, and stores them.

Not sure if this issue can be easily resolved though, due to how crafting works in 1.12 (completely server-side).

josephcsible:
https://github.com/CyclopsMC/CyclopsCore/blob/master-1.12/src/main/java/org/cyclops/cyclopscore/persist/world/GlobalCounters.java#L10 seems to indicate otherwise.

<blockquote><img src="https://avatars2.githubusercontent.com/u/11808375?s=400&v=4" width="48" align="right"><div><img src="https://assets-cdn.github.com/favicon.ico" height="14"> GitHub</div><div><strong><a href="https://github.com/CyclopsMC/CyclopsCore">CyclopsMC/CyclopsCore</a></strong></div><div>CyclopsCore - Minecraft coremod for EvilCraft, Integrated Dynamics and others.</div></blockquote>

rubensworks:
Not sure if that description is accurate, I don't remember ever implementing server-client syncing, and I also don't think vanilla does this out-of-the-box for WorldStorage.

Fatmice:
This remains a problem in 1.12.2-0.11.17

[...]

PR #430, opened 2017-12-01
Allow crafting two omni-directional connectors together to link them

josephcsible:


rubensworks:
How does this work? Doesn't this loose the ID? Or is an existing recipe handler automatically taking care of this?

Which ID will be assigned? The first or second one? Or a new one? (These things should probably be documented in the book as well)

josephcsible:
The existing recipe handler automatically takes care of it.

rubensworks:
Ok, great!

src/main/resources/assets/integrateddynamics/lang/en_us.lang
```diff
@@ -1360,6 +1360,7 @@ info_book.integrateddynamics.manual.parts.other.connector_omni_directional=Omni-
...
 info_book.integrateddynamics.manual.parts.other.connector_omni_directional.text2=You may for example want to connect more than two networks together using a single connector or connect with networks in other dimensions. For these cases, you can use the &lOmni-Directional&r connectors.
 info_book.integrateddynamics.manual.parts.other.connector_omni_directional.text3=These connectors will have a unique group id when first crafted, which identifies the group in which networks can be connected. Each connector group has a unique color of particles when placed in the world, the group id can be seen in the item tooltip or by right-clicking on the connector when placed. Additional connectors can be added to a group by crafting a new one using an existing connector within a group.
+info_book.integrateddynamics.manual.parts.other.connector_omni_directional.text4=You can also change the group of existing connectors in two ways. Crafting one alone in a crafting table will assign it to a new group with no other members, and crafting two together will copy the first one's group to the second one.
```
josephcsible, line 1363: It seems that the first half of this is a lie; you can't reset IDs like that. (It just looks like you can at first because of #429.)

PR #432, opened 2017-12-01
Use isRemote in place of isClientSide where possible

josephcsible:


rubensworks:
There were that many of them?! I thought I only used it when there was no other way...

src/main/java/org/cyclops/integrateddynamics/tileentity/TileVariablestore.java
```diff
@@ -106,7 +106,7 @@ public void onDirty() {
     @Override
     public void onLoad() {
         super.onLoad();
-        if(!MinecraftHelpers.isClientSide()) {
+        if(!world.isRemote) {
```
josephcsible, line 109: Ugh, apparently world can be null here.

PR #433, opened 2017-12-02
Revert "Use isRemote in place of isClientSide where possible"

josephcsible:
So apparently world can be null sometimes.

This reverts commit 1d9bede08ed77684a9a46a980b57bca3db7794eb.


rubensworks:
Wait, instead of reverting the whole commit, wouldn't it be better to just add null checks? If the world is null, none of that logic should run in any case.

josephcsible:
I haven't had time to give that as close of a look as I'd want to make sure that it doesn't introduce any subtle new bugs. Can you merge this as-is just to unbreak master, and then I'll revisit this at a later time?

PR #439, opened 2017-12-07
Give 3 omni-directional connectors instead of 2 when consuming one

josephcsible:
It isn't really fair to make the recipe effectively twice as expensive just
to join one to an existing pair.


rubensworks:
Sure, why not.

@josephcsible I thought the book mentioned something about this in the text, can you check if this needs updating as well?

rubensworks (approved):

josephcsible:
It looks like the book's text remains correct with this change.

PR #442, opened 2017-12-07
Fix IgnoredBlockStatus only handling valid enums

AlexIIL:
This catches the potential IllegalArgumentException that might get thrown if some-one passes something that isn't a valid enum value in Status, such as "hello" into the function. This now matches the behaviour of vanilla minecraft's PropertyEnum.parseValue, which stores it in an internal map (of whose get function never throws).

I'm not sure when this could happen in vanilla, however it looks like some commands and advancement criteria call this in order to parse their respective strings.

This was found with an in-dev version of buildcraft (https://gist.github.com/LuigiHutch/89da294073722bab676e2a77a69c63ea)

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/block/IgnoredBlockStatus.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/block/IgnoredBlockStatus.java
@@ -48,7 +48,11 @@ public class IgnoredBlockStatus extends IgnoredBlock {
 
         @Override
         public Optional<Status> parseValue(String value) {
-            return Optional.fromNullable(Status.valueOf(value.toUpperCase(Locale.ENGLISH)));
+            try {
+                return Optional.of(Status.valueOf(value.toUpperCase(Locale.ENGLISH)));
+            } catch (IllegalArgumentException e) {
+                return Optional.absent();
+            }
         }
 
         @Override


josephcsible:
Why is BuildCraft passing us invalid values? I'd like to see a vanilla command that can cause this.

AlexIIL:
Values *usually* come from a different source, such as player input for commands or loaded from config files. Vanilla handles invalid values cleanly (by returning Optional.absent()), it would be useful for other mods to also match vanilla's behaviour so that we don't have to wrap everything in ````try { ... } catch (Throwable ignored) {}````.

josephcsible:
Can you provide an example of a command that will cause this with no other mods installed?

AlexIIL:
Looks like commands have a big try catch around them for if something goes wrong, I'll test for advancement criterion next.

AlexIIL:
Looks like vanilla never crashes.

josephcsible:
My inclination is to say that BuildCraft needs to handle this then, but I'll let @rubensworks have the final say.

rubensworks:
Thanks for the PR @AlexIIL!

I partially agree with @josephcsible regarding the comment that BC should not pass invalid values at all. However, since the return type of that method is an Optional, it seems reasonable to make it swallow invalid value errors and return absent values, so I'll merge it.

@AlexIIL The title includes WIP, is there anything else you want to include in this PR? Or can this be merged?

rubensworks (approved):

AlexIIL:
I created it on github before testing (its a really small amount of changes). I should have removed it after I actually tested it though.

PR #447, opened 2017-12-14
Add string operators, localisation, tests.

josephcsible:
Updated version of #355 with the fixes discussed there.


gradle/forge.gradle
```diff
@@ -31,6 +31,11 @@ dependencies {
         exclude group: 'org.cyclops.cyclopscore', module: 'CyclopsCore'
     }
 
+    compile "com.google.re2j:re2j:1.1"
+
+    shadow("com.google.re2j:re2j:1.1") {
+    }
```
josephcsible: Is this the right way to do shading? It seems to work.

rubensworks: `shadow "com.google.re2j:re2j:1.1"` should work as well. Can you try if this works? Looks a bit cleaner :-)

rubensworks:
@tekacs Could you sign the CLA? (I thought you already did that in the other PR?...)

rubensworks:
How large is that shaded jar btw? Does it increase the resulting ID jar a lot?

josephcsible:
The jar I'm shading is only about 128kB.

josephcsible:
Two improvements I thought of before we merge this:

1. When we catch a PatternSyntaxException, we should give its details to the player.
2. We should make this curry in such a way that compiled regexes can be reused, instead of recompiling for each match.

rubensworks:
> When we catch a PatternSyntaxException, we should give its details to the player

+1. Just catch and rethrow as EvaluationException.

> We should make this curry in such a way that compiled regexes can be reused, instead of recompiling for each match.

Not sure about this. This would involve caching, and possibly elaborate cache invalidation schemes which may end up becoming less efficient than just recompiling. Let's avoid premature optimization.

josephcsible:
I actually wasn't thinking of caching at all. I was thinking of making it be partially applied such that when it gets the regex, it compiles it at that point, then the resulting operator carries the compiled regex in place of the string.

rubensworks:
So you want to make it so that players first have to compile strings to regexes before they can be used? Or am I not understanding you correctly?

I wouldn't do that, it goes at the cost of usability. Let's look into optimizations from the moment compilation actually seems to be a problem.

josephcsible:
Not quite. I mean that if you partially apply one of the functions and give it just a regex, it should compile the regex at that point.

EDIT: Thinking about it, there's no reason to hold up the merge for that. Let's just do it later.

PR #458, opened 2017-12-20
Make magma blocks dry into netherrack

josephcsible:


rubensworks:
Hmm, I'd rather not add this recipe to the regular drying basin. It doesn't make much sense, as it will already ignite when you insert lava into it. (Will merge this and rewrite the commit)

PR #470, opened 2017-12-29
Make aspects (and operators) only update on block updates

rubensworks:
This makes it so that certain aspect variables are not updated/invalidated every network tick anymore, but only when an actual block update is received.
These changes are then cascaded through to the operators.

Operations on fully static variables will now also not be invalidated every tick anymore.

@josephcsible This change may potentially break a lot of things, so I need a second set of eyes on this. Could you check if I didn't miss anything that could break with this change?

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java
+++ b/src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java
@@ -39,6 +39,7 @@ import org.cyclops.integrateddynamics.api.evaluate.variable.IValue;
 import org.cyclops.integrateddynamics.api.network.IEnergyConsumingNetworkElement;
 import org.cyclops.integrateddynamics.api.part.PartPos;
 import org.cyclops.integrateddynamics.api.part.PartTarget;
+import org.cyclops.integrateddynamics.api.part.aspect.AspectUpdateType;
 import org.cyclops.integrateddynamics.api.part.aspect.IAspectRead;
 import org.cyclops.integrateddynamics.api.part.aspect.IAspectRegistry;
 import org.cyclops.integrateddynamics.api.part.aspect.IAspectWrite;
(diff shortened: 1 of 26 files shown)


josephcsible:
I haven't forgotten about this; I'll try to get to it soon. (I note that it has a merge conflict right now, though.)

src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java
```diff
@@ -122,7 +129,8 @@ public static void load() {}
                             // Catch possible errors
                         }
                         return null;
-                    }).handle(AspectReadBuilders.PROP_GET_NBT, "tile").buildRead();
+                    }).withUpdateType(AspectUpdateType.BLOCK_UPDATE)
```
josephcsible: Doesn't NBT frequently change even without a block update?

src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java
```diff
@@ -538,22 +547,27 @@ public static void load() {}
...
             public static final IAspectRead<ValueTypeBoolean.ValueBoolean, ValueTypeBoolean> BOOLEAN_CLOCK =
                     AspectReadBuilders.Redstone.BUILDER_BOOLEAN_CLOCK.handle(AspectReadBuilders.PROP_GET_BOOLEAN, "clock").buildRead();
 
             public static final IAspectRead<ValueTypeInteger.ValueInteger, ValueTypeInteger> INTEGER_VALUE =
-                    AspectReadBuilders.Redstone.BUILDER_INTEGER.handle(AspectReadBuilders.PROP_GET_INTEGER, "value").buildRead();
+                    AspectReadBuilders.Redstone.BUILDER_INTEGER.withUpdateType(AspectUpdateType.BLOCK_UPDATE)
+                            .handle(AspectReadBuilders.PROP_GET_INTEGER, "value").buildRead();
             public static final IAspectRead<ValueTypeInteger.ValueInteger, ValueTypeInteger> INTEGER_COMPARATOR =
-                    AspectReadBuilders.Redstone.BUILDER_INTEGER_COMPARATOR.handle(AspectReadBuilders.PROP_GET_INTEGER, "comparator").buildRead();
+                    AspectReadBuilders.Redstone.BUILDER_INTEGER_COMPARATOR.withUpdateType(AspectUpdateType.BLOCK_UPDATE)
```
josephcsible, line 586: Can't this also change without a block update?

src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java
rubensworks: True, this was part of a small test I did, but should be removed indeed.

src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java
rubensworks: It should only change with neighbour changes, see the code in `TileEntity#markDirty`.

Issue #195, opened 2017-06-17
Feature Request: Functions


ricochet1k:
So, I'm trying to come up with a way to use Integrated Dynamics and Tunnels to make an autocrafting system for Botania. Here's what I've come up with so far, in pseudo-Haskell style:
```haskell
targetList :: List Item
targetList = <proxy>

currentInv :: List Item
currentInv = <chest inventory>

stillNeed :: List Item
stillNeed = filter (not . haveEnough) targetList

haveEnough :: Item -> Boolean
haveEnough item = gt (size item) . haveAlready item

haveAlready :: Item -> Int
haveAlready = itemCounts . sameItems

itemCounts :: List Item -> Int
itemCounts = reduce (+) . map size

sameItems :: Item -> List Item
sameItems = (flip filter) currentInv . eq
```

The idea is to have an inventory reader and an item exporter attached to a chest, with an item interface somewhere else. `currentInv` would be the Inventory Items aspect of the inventory reader, targetList would be a proxy where you can insert the list of items you want placed in the chest, exactly, and no more.

My Haskell is a bit rusty, but I think this would work. Maybe there's a different way to do this, but as far as I can tell, it isn't possible to do the `haveEnough` part because the `item` parameter is used twice.

So, my request is this: Let us make functions. Let us make Parameter cards, and then make a function by inserting the Parameter cards and the result card. Parameter cards can be duplicated and then used in multiple places. Also, if the result card refers to a Parameter card that's not included in the function, it should still be able to be used inside another function. Or in other words, don't prevent constructing the function if all the Parameters are not included.

Functions would also be useful to make composing Operators easier. Point-free style is not intuitive for most people.

rubensworks:
CyclopsMC/IntegratedTunnels#37 is a similar issue where more complex parameter handling is required.

I'll have to think about this a bit, to see what will be the best solution given the way ID currently works.

josephcsible:
We don't actually need user-defined functions to do what you need; we just need the S combinator (`S = λxyz.xz(yz)`, called `ap` in Haskell):
```haskell
haveEnough :: Item -> Boolean
haveEnough = ap (gt . size) haveAlready
```
Proof:
```
haveEnough = S (B gt size) haveAlready
haveEnough item = S (B gt size) haveAlready item
haveEnough item = (B gt size) item (haveAlready item)
haveEnough item = B gt size item (haveAlready item)
haveEnough item = gt (size item) (haveAlready item)
```
[...]

Issue #352, opened 2017-11-08
We need to do the opposite of currying


josephcsible:
Currently, we perform currying, where if you call a function with less arguments than it needs, it gets partially applied. We need to do the opposite of that as well, where if you call a function with too many arguments, the remaining ones get passed to the function's output, transforming f(a, b) to (f(a))(b) for example.

Here's a concrete case that doesn't work as expected:
1. Create variable cards with the number 3, the Increment operator, and the Identity operator.
2. Go to Flip in the logic programmer, put in Identity, and get the resulting card (I'll call it "CI").
3. Go to Apply 2 in the logic programmer, put in CI, 3, and Increment, and get the resulting card.
4. Put that card in a display panel.

The expected result is the number 4, but instead you get an error: "The given config render pattern with 2 slots is not compatible with the number of input types 1 for :flip".

Here's what I expect to happen, written in combinatory logic style:
* Flip Identity 3 Increment
  * Apply beta reduction to Flip (λxyz.xzy):
* Identity Increment 3
  * Apply beta reduction to Identity (λx.x):
* Increment 3
  * Evaluate the Increment operator:
* 4

(Edit: In fact, you get the error much sooner by just attempting to materialize CI.)

rubensworks:
Ok, I understand the problem. This is indeed something I did not take into account while implementing currying. But AFAICS, it should be possible to implement this in the current system without any major changes nonetheless.

josephcsible:
In case anyone else is experiencing this issue, a workaround is to use "Apply" (or "Apply2" or "Apply3" in case of really complicated examples) in place of "Identity". This won't change the meaning of your expression, since they're equivalent up to eta-conversion.

PR #506, opened 2018-01-27
Add the W combinator

josephcsible:
Fixes part of #195.

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/CombinedOperator.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/CombinedOperator.java
@@ -14,6 +14,7 @@ import org.cyclops.integrateddynamics.api.evaluate.variable.IValueType;
 import org.cyclops.integrateddynamics.api.logicprogrammer.IConfigRenderPattern;
 import org.cyclops.integrateddynamics.core.evaluate.variable.ValueHelpers;
 import org.cyclops.integrateddynamics.core.evaluate.variable.ValueTypeBoolean;
+import org.cyclops.integrateddynamics.core.evaluate.variable.ValueTypeOperator.ValueOperator;
 import org.cyclops.integrateddynamics.core.evaluate.variable.ValueTypes;
 
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/CombinedOperator.java
```diff
@@ -253,6 +254,54 @@ public CombinedOperator newFunction(IOperator... operators) throws EvaluationExc
...
+                return ValueHelpers.evaluateOperator(operator, values);
+            }
+        }
+
+        public static CombinedOperator asOperator(IOperator operator) throws EvaluationException {
+            CombinedOperator.Duplicate duplicate = new CombinedOperator.Duplicate(operator);
+            IValueType[] duplicatedInputTypes = new IValueType[] {operator.getInputTypes()[0]};
+            try {
+                return new CombinedOperator("W", "duplicated", duplicate, duplicatedInputTypes,
+                        operator.getOutputType(), IConfigRenderPattern.PREFIX_1);
```
josephcsible: I'm worried that `operator.getOutputType()` may be incorrect in the case that uncurrying happens.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/CombinedOperator.java
```diff
@@ -253,6 +254,54 @@ public CombinedOperator newFunction(IOperator... operators) throws EvaluationExc
...
+                return ValueHelpers.evaluateOperator(operator, values);
+            }
+        }
+
+        public static CombinedOperator asOperator(IOperator operator) throws EvaluationException {
+            CombinedOperator.Duplicate duplicate = new CombinedOperator.Duplicate(operator);
+            IValueType[] duplicatedInputTypes = new IValueType[] {operator.getInputTypes()[0]};
+            try {
+                return new CombinedOperator("W", "duplicated", duplicate, duplicatedInputTypes,
+                        operator.getRequiredInputLength() == 1 ? ValueTypes.CATEGORY_ANY : operator.getOutputType(), IConfigRenderPattern.PREFIX_1); // TODO output type could be more specific when uncurrying
```
josephcsible, line 285: Is it okay for the output type of an operator to just be CATEGORY_ANY? If not, I don't see a good way to do uncurrying without rewriting our type system.

josephcsible:
This is actually kind of redundant to #509, and in any case, I don't think it should merge now until #352 is solved.

Issue #519, opened 2018-01-31
Can't change properties of erroring aspect


josephcsible:
Suppose you make this change:
```diff
diff --git a/src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java b/src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
index 6d4ab93..445bd89 100644
--- a/src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
+++ b/src/main/java/org/cyclops/integrateddynamics/part/aspect/read/AspectReadBuilders.java
@@ -26,6 +26,7 @@ import org.cyclops.commoncapabilities.api.capability.work.IWorker;
 import org.cyclops.cyclopscore.datastructure.DimPos;
 import org.cyclops.cyclopscore.helper.TileHelpers;
 import org.cyclops.integrateddynamics.Capabilities;
+import org.cyclops.integrateddynamics.api.evaluate.EvaluationException;
 import org.cyclops.integrateddynamics.api.evaluate.variable.IValue;
 import org.cyclops.integrateddynamics.api.network.IChanneledNetwork;
 import org.cyclops.integrateddynamics.api.network.INetwork;
@@ -340,6 +341,7 @@ public class AspectReadBuilders {
             DimPos dimPos = input.getLeft().getTarget().getPos();
             INetwork network = NetworkHelpers.getNetwork(dimPos.getWorld(), dimPos.getBlockPos());
             int channel = input.getRight().getValue(PROPERTY_CHANNEL).getRawValue();
+            if(channel == -2) throw new EvaluationException("test");
             return network != null && network.hasCapability(EnergyNetworkConfig.CAPABILITY) ? network.getCapability(EnergyNetworkConfig.CAPABILITY).getChannel(channel) : null;
         };
 ```

If you set a network reader's channel to -2, as expected, the display changes to "ERROR". If you try to change the channel back to anything else, the error doesn't go away. It's stuck until you do something drastic like saving and reloading. (This actually came up in something more complicated, but this example makes the problem a lot simpler to see.)

PR #521, opened 2018-01-31
Let parts expose more than one value

josephcsible:


josephcsible:
This should not be merged until #519 is fixed.

rubensworks:
On hold until we have looked into #524.

rubensworks:
@josephcsible I assume this can be merged now, now that #524 is resolved?

josephcsible:
I thought the intent was that #524 would make doing this unnecessary.

rubensworks:
I don't remember what we discussed (I don't think it was here on GitHub, right?). But I think this issue is orthogonal to #524, and doesn't solve the same problem.

Issue #348, opened 2017-11-06
Variables should be visible in all parts, not just variable stores.


josephcsible:
Currently, only variable stores are checked for variable cards when searching the network for them. Other parts, such as writers, display panels, proxies, and materializers, should also be searched, to avoid occasionally needing to copy a variable card for no other reason. Originally reported in the comments of #295.

rubensworks:
I think I looked into this at one point, but the implementation was non that trivial. Perhaps something for the future.

rubensworks:
For reference: https://github.com/CyclopsMC/IntegratedDynamics/commit/7bb04f5200dfddbe6b5041361cf5f6a9a7518b36 changes some things related to variable references in networks, which may make it possible for non-variable stores to act as variable store.

tdaffin:
I've been working on a fix for this on my fork and I have a version which works up to a point: https://github.com/tdaffin/IntegratedDynamics/commit/590105dbe9c316da9b65a4af995d7e7e10689a3b

The problem is that if you have an operator variable that depends on a couple of other operator variables that are themselves in parts then it does not get updated.  I figure I need to something more clever with the sendVariablesUpdateEvent flag to fix that.

I figured I'd post here in case it helped, or you guys have any suggestions on how I should proceed, or if perhaps I'm just totally missing the point ;)

tdaffin:
I've pushed a few commits to my development branch that fix the problems with operators in parts not being seen by operators that depend on them.
A comparison of my entire branch with master is here: https://github.com/CyclopsMC/IntegratedDynamics/compare/master-1.12...tdaffin:part-variables

A new concern at this point is whether the call to onVariableContentsUpdated in PartTypeWriteBase will cause performance issues?

rubensworks:
@tdaffin As far as I can see at the moment, you're approach seems like a good solution. Making part states also have the variable container cap seems like the logical thing to do.

The only concern I have atm are the additional calls to ' onVariableContentsUpdated' in the parts (as you have mentioned as well), which can be expensive. This should not be called when a variable updates its value, instead, this should be called when a different variable/aspect becomes activated. Could you have a look to see if these calls can be avoided? Because at the moment, this causes _all_ variables to become updated in the network _every tick_, which will be quite inefficient.

(Please correct me if I misunderstood anything, as I browsed through the code rather quickly, I'll do a more thorough review when you submit your PR)

tdaffin:
[...]

Issue #545, opened 2018-02-19
Relational Operators of live-updating Block-NBT values does not update.


Timetech01:
#### Issue type:

_Only keep the issue types that are applicable._

- :bug: Bug

____
#### Short description:

Attempting to control networks and general Minecraft systems using Relational Operators that receive at-least one of their input values from actively updating NBT values causes invalid behavior.

#### Expected behaviour:

When BurnTime would become less than 50 or equal-to-or-greater-than 50, I would expect the Redstone Writer and its monitoring Display Panel to switch between True and False or vise-versa, depending on the current state of BurnTime.

#### Actual behaviour:

Once the Less Than Operator has determined its current Boolean value, it does not update unless a change to the components of the network is made, apparently forcing a refresh of the system.

#### Steps to reproduce the problem:

Using the Block Reader on a Furnace, I gained the Tile Entity NBT.
I then found the NBT Value Integer of the "BurnTime" String inside the Tile Entity NBT data.
I then used the Relational Operator < "Less Than" to compare the current "BurnTime" Integer to the constant Integer 50.
I then put this Relational Operator in the Redstone: Boolean slot of a Redstone Writer, and a copy of it inside a Display Panel for convienence.  [I was also tracking the entire NBT of the Furnace, as well as its BurnTime on Display Panels.]

____
#### Versions:

_Exact versions of the following mods, not just *latest*._

- This mod: 1.12.2-0.11.5
- CyclopsCore: 1.12.2-0.11.4
- Minecraft: 1.12.2
- Forge: 14.23.2.2618

#### Log file:

https://pastebin.com/2046uYUX

Log File edited to preserve anonymity.

rubensworks:
Note to self: similar to #539.

rubensworks:
I can't seem to reproduce this myself, I also tried similar variants of your described steps.

Are you sure that you're using version 1.12.2-0.11.5?
If yes, could you describe a detailed step-by-step guide to consistently reproduce the problem? (please double check if you can reproduce the problem yourself using these steps)

(If possible, a video showing the problem and how to set it up would be great!)

rubensworks:
@Timetech01 Any updates on this?

Timetech01:
@rubensworks Working out how to record video and audio of my setup.

Timetech01:
@rubensworks https://www.youtube.com/watch?v=oM2MtlysFoM

rubensworks:
[...]

PR #547, opened 2018-02-20
Make variables visible in all parts, not just variable stores

tdaffin:
This is intended to close issue #348.

Other than making part states also have the variable container capability I had to fire VariableContentsUpdatedEvent in the following places:
* PartTypeWriteBase#updateActivation
* AspectWriteBase#update -- but only when the aspect is activated or deactivated
* PartTypeVariableDriven#onDirty

As far as I can see none of the places fire the event every tick, and manual testing of various networks in a test world seem to work well.


rubensworks:
Thanks! I'll try to review it this week.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetwork.java
```diff
@@ -127,6 +129,13 @@ public boolean hasPart(int partId) {
                     it.remove();
                 }
             }
+            // Also check parts
+            for(PartPos partPos : partPositions.valueCollection()){
```
rubensworks: `for(PartPos partPos : partPositions.valueCollection()) {`

rubensworks: `for(PartPos partPos : partPositions.valueCollection()) {`

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetwork.java
```diff
@@ -127,6 +129,13 @@ public boolean hasPart(int partId) {
...
+                IPartContainer partContainer = PartHelpers.getPartContainer(partPos.getPos(), partPos.getSide());
+                IVariableContainer variableContainer = partContainer.getCapability(VariableContainerConfig.CAPABILITY, partPos.getSide());
+                if (variableContainer != null)
```
rubensworks: Add `{` ... `}`

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetworkElement.java
```diff
@@ -31,7 +32,7 @@
  */
 @Data
 public class PartNetworkElement<P extends IPartType<P, S>, S extends IPartState<P>> extends NetworkElementBase
-        implements IPartNetworkElement<P, S>, IEnergyConsumingNetworkElement {
+        implements IPartNetworkElement<P, S>, IEnergyConsumingNetworkElement, IEventListenableNetworkElement<P> {
```
rubensworks: Not needed here, `IPartNetworkElement` already implements this.

src/main/java/org/cyclops/integrateddynamics/core/part/panel/PartTypePanelVariableDriven.java
```diff
@@ -144,6 +144,9 @@ public void update(INetwork network, IPartNetwork partNetwork, PartTarget target
             // tick-1: Part tick: update the value again, the old value has still not been sent here!
             // tick-1: -- send all block updates to client --- This will contain the value that was set in tick-1.
             state.onDirty();
+            if (network != null) {
+                network.getEventBus().post(new VariableContentsUpdatedEvent(network));
```
rubensworks: Placing this call here will cause panels to trigger `VariableContentsUpdatedEvent` every time the displayed value changes. This means that showing something like the world time will become _very_ expensive.

src/main/java/org/cyclops/integrateddynamics/part/aspect/write/AspectWriteBase.java
```diff
@@ -61,6 +66,14 @@ public AspectWriteBase(ModBase mod, ModBase modGui, String unlocalizedTypeSuffix
             }
         } else if(!writerState.isDeactivated()) {
             onDeactivate(partTypeWriter, target, writerState);
+            postVariableContentsUpdatedEvent(target.getCenter());
+        }
+    }
+
+    private static void postVariableContentsUpdatedEvent(PartPos partPos){
```
rubensworks: Is this event call really needed again here? Are there cases where the call from `PartTypeWriteBase` is not sufficient? If so, do you think it would be possible to combine them somehow?

rubensworks (changes requested):
Most things look good. Just some comments regarding when the variable update event is called.

src/main/java/org/cyclops/integrateddynamics/part/aspect/write/AspectWriteBase.java
tdaffin: No, I don't think it is necessary -- I think perhaps I thought it was at some point because I had a network that was in a weird state due to some other stuff I'd been doing.
I've removed it.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetworkElement.java
tdaffin: Removed.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetwork.java
tdaffin: Done

src/main/java/org/cyclops/integrateddynamics/core/part/panel/PartTypePanelVariableDriven.java
tdaffin: This is the tough one...
I've removed this call and instead added a call in ContainerPartDisplay#onDirty which prevents the performance problem but also detects changes when display inventories are changed.
The remaining problem then is that if you have a display that is displaying the result of an operator that combines the values from two other displays then it does not update.
This screenshot shows a test network I've been using:
image
The bottom row of displays contain redstone cards from the readers below, lets say they have ids 1, 2, 3, 4 in that order.
The second row up has variable cards that are set by 'anding' together the following variables from below (1 && 2, 1 && 2, 3 && 4, 3 && 4), and they have ids 5, 6, 7, 8.  This means that if you switch off either of the two switches on the left then both the left hand lamp and left hand display on the second row should go off and false.  Similar on the right.  Even with removing this call, this row still works fine.
The problem occurs in the third row, where the top two displays are made by (6 && 7, 5 && 8), i.e the left hand display is the two cards from the displays on the second row 'anded' together and the right hand display is the two cards from the writers on the second row 'anded' together.
This means that if any of the switches are off then both top displays should be false, and if all are on then both should be true.
The problem is that the top left display (cards from displays combined in an operator) does not update when I flip one of the switches on the second row.
I'm going to go ahead and push the latest version so we can discuss how to proceed.
If you need a test world with the setup in I can supply that.  I'm also working on getting an integration test setup with the network, but that is going to take time :)
On that note I hope that my attempts here aren't taking up too much of your time -- I'm kind of using this as a way to learn MC modding in general...

tdaffin:
To make it clear the code as currently posted has one remaining problem:

The problem is that if you have a display that is displaying the result of an operator that combines the values from two other displays then it does not update.
This screenshot shows a test network I've been using:
image
The bottom row of displays contain redstone cards from the readers below, lets say they have ids 1, 2, 3, 4 in that order.
The second row up has variable cards that are set by 'anding' together the following variables from below (1 && 2, 1 && 2, 3 && 4, 3 && 4), and they have ids 5, 6, 7, 8.  This means that if you switch off either of the two switches on the left then both the left hand lamp and left hand display on the second row should go off and false.  Similar on the right.  Even with removing this call, this row still works fine.
The problem occurs in the third row, where the top two displays are made by (6 && 7, 5 && 8), i.e the left hand display is the two cards from the displays on the second row 'anded' together and the right hand display is the two cards from the writers on the second row 'anded' together.
This means that if any of the switches are off then both top displays should be false, and if all are on then both should be true.
The problem is that the top left display (cards from displays combined in an operator) does not update when I flip one of the switches on the second row.
I'm going to go ahead and push the latest version so we can discuss how to proceed.
If you need a test world with the setup in I can supply that.  I'm also working on getting an integration test setup with the network, but that is going to take time :)

tdaffin:
I've been debugging into the remaining problem this morning, and so far the best I have is that when the variables that are in the middle displays in the second row (6 and 7) are invalidated they are not also invalidating the variable that depends on them in the top row.
I'm looking into why that is now, though I suspect that the VariableAdapter#dependents list is empty when VariableAdapter#invalidate is called, though my other debug code indicates that it is being populated when LazyExpression#evaluate is called on the variable in the top row (I'll call this 9).

tdaffin:
Ok, I think I've figured it out.
The problem is that PartStateActiveVariableBase#getVariable is still creating a new IVariableFacade where it should be referencing the one that is in it's new variableContainer.
I'll clean up my fix and update the PR in a bit.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetwork.java
tdaffin: Addressed below

tdaffin:
Ok, I think I have it working as it should now.  All my manual testing worked as expected.  I haven't got an integration test for this working yet, but I think that may take a while.

src/main/java/org/cyclops/integrateddynamics/core/network/PartNetwork.java
tdaffin: I think I addressed this below, or did you mean to say something else about this?

tdaffin: Ah, finally spotted it -- there needed to be a space before the brace.
Fixed!

rubensworks:
Hmm, just wondering if the last problem you described might be related to #545? (which I cannot reproduce myself)

> I haven't got an integration test for this working yet, but I think that may take a while.

That's not needed in this case. I mainly write those for pure evaluation stuff.

So, is this good to review again?

tdaffin:
I wondered when I found the root cause of the problem if the fix might fix some other things, but it is a pretty vague hunch more than a sure feeling.

In case it helps the observation that lead me to figure out how to fix it was that I had more than one instance of LazyExpression with the same id.  I logged out the id and the System.identityHashCode(this) to figure out what was going on.

The code is ready for review again.

rubensworks (approved):
Ok, looks great, thanks! Will merge this at the next feature release.

rubensworks:
@tdaffin Oh, one more small question: did you test if this works properly in combination with proxies/materializers/delays? Those do some special things, which _may_ potentially break this.

I'll test all of this myself in any case when I prepare the next feature release.

tdaffin:
Awesome -- this is some functionality I've been wanting since I first used ID.

I had not done any testing with proxies, materializers or delayers, but I did this morning and things seem to be working ok.  
My testing was not super extensive, mainly focusing on things I could do both with and without these changes.  
I realize now that I haven't really looked at whether I can reference a variable that references a proxy or delayer if it is in a part such as a writer or display.  I'll go give that a quick try now.

tdaffin:
I tested various variables made by referencing proxies and delayers, and operations on them, all with the variables only in displays instead of in a variable store and all worked as I expected.

rubensworks:
For reference, I didn't forget about this. The plan is to merge this before the next big release in a couple of weeks/months.

tdaffin:
Great news.  I have been busy with other things in life since I did this, but it will be nice to see it released.

tdaffin:
I've merged the latest master-1.12 to fix the conflicts and done a very quick in game test of the the worlds I was using before.
Everything seemed to work ok, but I could have easily missed some detail as it has been a long time since I looked at this.
Are you intending to also include the [feature-recipies](https://github.com/CyclopsMC/IntegratedDynamics/commits/feature-recipes) branch in the next release?  If so I could do some local testing to see if it will merge ok with these changes to give time to fix any problems.

rubensworks:
Ok, thanks.

No need to check with feature-recipes, I'll do that when I merge :-)

rubensworks:
Thanks a lot again for the PR, will prepare a release in a week or so.
And sorry for the long delay, this update has taken longer than I wanted...

Had to rebase and squash a bit, so merged this as affaab44aa342b1865882af4674375cefe8dfd87.

PR #553, opened 2018-02-22
Fix a couple of integration tests

tdaffin:
I've been looking into what it would take to add some integration tests in order to test my changes for https://github.com/CyclopsMC/IntegratedDynamics/pull/547 and have discovered that four of the 344 of the integration tests were already failing.
In this PR I'm fixing two of the failing tests that were pretty obviously wrong or in need of an update.


rubensworks:
I recently fixed this in another branch as well, but It'll take a while before that branch will be merged, so I'll merge your fixes already. Thanks!

tdaffin:
Great!  Do you have fixes for the other two failing tests already?

They are:

-  Test TestItemStackOperators#testItemStackOreDictStacks failed!
     java.lang.AssertionError: Failure: size(oredict_stacks(plankWood)). Expected 2, but got 7.
 
 
-  Test TestItemStackOperators#testItemStackStoredFe failed!
     java.lang.AssertionError: Failure: storedfe(energyBatteryFull) == 100000. Expected 100000, but got 2000.

I haven't figured out why the first one is failing yet, but the second appears to be due to 8b77532e1fb611fd4532efd944b7d8a62abaded0 though I'm not sure of they right way to fix it yet.

rubensworks:
The first one was just due to other mods being loaded in my dev environment. But your fix should be correct in any case.

PR #554, opened 2018-02-22
Fix another integration test

tdaffin:
This pull request fixes one of the two remaining failing tests from https://github.com/CyclopsMC/IntegratedDynamics/pull/553

The problem was that the test was using the standard ore dictionary call but the operator was using a method where the wildcard item was expanded out into the individual items from the creative tab.

To fix the problem I moved the code that does the expansion into a helper function so that the test can use it too.


rubensworks:
Great, thanks!

Issue #616, opened 2018-07-28
Flip operator with 3+ inputs


vitalexbel:
#### Issue type:

- :heavy_plus_sign: Feature request

____

#### Short description:

Flip operator to flip operator with 3 inputs.
For example, `reduce` operator has 3 inputs (operator, list, initial value).
And I want to apply `initial value` and `operator` first, and then get an operator with only `list` input needed.

So I want to get this order of input values: 312

Maybe you can make an operator that can rearrange 3rd input value to to the beginning
Thanks

bsngardner:
This is definitely needed. You might call it rotate.

bsngardner:
I don't know if you had any specific plans for this, but I implemented it. Is it okay if I submit a pull request?

josephcsible:
You don't need another operator for this; it's already possible. In Pipe, put in the Reduce and Flip operators, then in Flip, put the result of the Pipe. The output will be Reduce, but with argument order "initial value", "operator", "list", as you want.

Here's a proof of how/why this works, in lambda calculus/combinatory logic notation (flip = C and pipe = (C B)):
* flip (pipe reduce flip)
* \x.\y.flip (pipe reduce flip) x y
* \x.\y.pipe reduce flip y x
* \x.\y.flip (reduce y) x
* \x.\y.\z.flip (reduce y) x z
* \x.\y.\z.reduce y z x

bsngardner:
I'm not able to do this with variable cards, so I'm likely misunderstanding you. 
flip(pipe(flip, reduce)) is what I did. The result is List->operator->any->any. 
In case I misunderstood the order, I tried flip(pipe(reduce,flip)) (this makes no sense to me, as flip takes an operator and reduce outputs type Any from whatever comes from the list). The result is List->operator->Any->operator.  Though I don't know what it would do. 

I'm not the OP, but I guess I am hijacking this ticket. I'm looking for a more generic way to get later inputs to the front, since my requirement is different from the OP's. I'm looking at lambda calculus to try to understand what you put there. Does "\\" there mean lambda or something else?

rubensworks:
Sure, it may already be possible, but it's not trivial to do. So I'm happy with the PR on this.

Closing this as this will be available soon because of PR #636.

PR #636, opened 2018-10-22
Add Rotate operator

bsngardner:
rotates the inputs to an operator, similar to flip

I don't have chinese lang stuff because I don't know chinese. I hope I didn't miss anything; I based it on the flip operator pretty heavily.


rubensworks:
Thanks! I'll review this in a week or two (on holidays right now)

bsngardner:
I had a thought yesterday: what do you think of making the operator take an integer input to rotate the inputs N times? That way you could get any input to the front with one card. Just a thought; this new operator works to make it possible to reach the later inputs without consuming the first ones.

rubensworks:
> what do you think of making the operator take an integer input to rotate the inputs N times?

That makes sense! Feel free to include it in this PR.

For usability, it might be good to keep the normal `rotate`, and introduce a new `rotate-n` or something like that.

josephcsible:
I don't think I like this. It breaks the idea that f(x, y) is basically equivalent to f(x)(y). It can cause visible differences and strange behavior when used in conjunction with already partially-applied and curried operators, or if an operator returns another operator.

Clarification: my issue with this is that it peeks at the number of operands that the current operator takes. If that value were instead provided to it by the player, or hardcoded to 3, I'd have no problem with this.

rubensworks:
@josephcsible Can you give a concrete example where you feel this will behave strangely?

I don't see any problems with this PR myself atm.

bsngardner:
I commented on Issue #616 about my attempts to follow @josephcsible's explanation on how to rotate operator inputs with the current feature set. If there is a generic way to do it now, cool. I don't actually understand the lambda calculus constraints mentioned, but hopefully we can figure out the best and proper way to have this feature.

bsngardner:
Also, does the flip operator not break the same idea, if I understand what you're saying?

rubensworks (approved):

rubensworks:
Looks good to merge for me.

@josephcsible Please let me know if you still have issue with this, and if so, please give a concrete example where you feel this will behave strangely

@bsngardner I'm still open to the idea of adding a `rotate-n` as well. Feel free to add it to this PR as well, otherwise I'll merge this PR as-is tomorrow.

bsngardner:
Okay, I'll write a rotate-n this evening

josephcsible:
This kind of fell off my radar. I'm taking another look now to come up with the concrete example.

josephcsible:
Here's one example: Create a list containing the operators Arithmetic Addition and Arithmetic Subtraction. Use "Apply" to apply that list to the "List Get" operator. The result of that is an operator with signature Integer -> Integer -> Integer -> Integer. Passing in 0, 20, and 15 to that operator will return 35, and passing in 1, 20, and 15 will return 5. (Note that if you want to try this, you need to use "apply" followed by "apply2" rather than just using "apply3" to do them all at once, due to bug #352.)

If I rotate that operator, I'd expect that passing 15, 0, and 20 would return 35, and that passing 15, 1, and 20 would return 15. What actually happens is that I get an error: "The operator Rotate received the operator Get with input length 1 while length 2 is required."

bsngardner:
Hmm and this is nontrivial to solve?

rubensworks:
@josephcsible That sounds like a bug, but I don't think the cause lies within this PR. I think the problem might be related to #352 even.

josephcsible:
Okay, here's a better example: imagine instead of `+` and `-`, the list contained `+` and the number 3. Now the arity of the rotated get operator depends on its own inputs in a circular, non-resolvable way.

rubensworks:
Interesting case, I wonder if/how this would crash...

Won't the same problem occur for the flip operator as well?

josephcsible:
No. Flip always just flips the first two operands to an operator, no matter how many there are (with the exception of bug #352). Granted, you'll probably end up trying to apply 3 as if it were an operator, but that's not flip's fault (and choosing a unary operator would make that problem go away too). The problem with this implementation of rotate is that it tries to figure out how many there are and then rotate them all, which can lead to paradoxes. I think the simplest solution is either to make it always assume 3 operands, or have the player pass the number of operands.

rubensworks:
I still don't understand the problem exactly tbh. I should sit down and play with it a bit I guess (but I don't have time for that unfortunately)

@bsngardner Could you have a look to see what happens at runtime when you try out @josephcsible's use cases?

If it indeed fails in some way, we should look into coming up with a different solution, or one of those suggested by @josephcsible.

> I think the simplest solution is either to make it always assume 3 operands, or have the player pass the number of operands.

I don't have a strong opinion on either of them.

josephcsible:
So I thought about this some more, and this isn't a good idea:

> what do you think of making the operator take an integer input to rotate the inputs N times?

Doing this would make it impossible to properly type-check this, since the overall type would be different depending on the values of its inputs, rather than just the types of its inputs.

> have the player pass the number of operands

Ditto.

For evidence of this, consider trying to build the function in Haskell. Rotating three arguments by one place is easy:
```
:t \f x y z -> f y z x
\f x y z -> f y z x :: (t3 -> t2 -> t1 -> t) -> t1 -> t3 -> t2 -> t
```
Even simpler than rotating an arbitrary number of places would be rotating either one or two depending on a Boolean, like this: 
```
:t \b f x y z -> if b then f y z x else f z x y
\b f x y z -> if b then f y z x else f z x y :: Bool -> (t1 -> t1 -> t1 -> t) -> t1 -> t1 -> t1 -> t
```

Note that x, y, and z now all have to be the same type. This isn't true for reduce (or basically any other function in ID), so such an operator would be useless.

bsngardner:
So what operator should we make? rotate-n is unnecessary, so it can be thrown out. I like the idea of passing how many operators should be included in the rotate; it's even more flexible. Are there any problems with this operator? rotate Int -> Op -> Op
I was able to accomplish what I needed this feature for without it using some reordering and meta operators (piping a pipe op to another operator or something, filtering a list and using the length to select an operator in another list, etc; harder but kinda fun). But I still think this would be a useful feature.

josephcsible:
> I like the idea of passing how many operators should be included in the rotate; it's even more flexible.

The problem is that such an operator can never be well-typed. In order for it to be well-typed without requiring all of the inputs to be of the same type (which would break your use case and pretty much every other use case), both the number of operands and the number of places to rotate must be fixed.

There is still a useful Rotate operator to be made, though. It's defined as `rotate f x y z = f y z x`. Just implement that. It's type-safe and works perfectly for your use case.

moonheart08:
Any status on this PR? This operator would be pretty helpful.

rubensworks:
@moonheart08 @bsngardner has been inactive, so feel free to take this PR over if you want.

moonheart08:
@rubensworks I'll consider it. (As in, I'll clone the repo right now, possibly do it, possibly not.)

bsngardner:
I haven't had time to change the rotate to just rotate 3 inputs instead of all of them. I believe that's what was decided. If you have time, it's a small change from what I put in this PR. I just don't have the IDE set up anymore.

rubensworks:
Closing due to inactivity. Feel free to submit a new PR.

PR #638, opened 2018-10-28
Only show the warning about copying for variable cards that are actually unsafe to copy

josephcsible:


rubensworks:
We could go one step further and also void the warning on constants without variable references.

But this is a good step in the right direction already :-)

josephcsible:
Can we do that? Wouldn't copying them still create two cards with the same ID, and wouldn't reprogramming one of them later then create two different variables with the same ID?

rubensworks:
Yes, you're right! I clearly didn't think it through :-)

PR #642, opened 2018-11-04
Fix #524: Support heterogeneous lists

### LINKED ISSUE #524: Support heterogeneous lists
#### Issue type:

_Only keep the issue types that are applicable._

We should support lists that contain values of different types.

josephcsible:


src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyMaterializedFactory.java
```diff
@@ -52,12 +57,20 @@ public String serialize(ValueTypeListProxyMaterialized<IValueType<IValue>, IValu
...
-        String[] values = new String[split.length - 1];
-        System.arraycopy(split, 1, values, 0, split.length - 1);
+        boolean heterogeneous = valueType.isCategory();
+        IValueType<IValue> elementValueType = valueType;
 
         ImmutableList.Builder<IValue> builder = ImmutableList.builder();
-        for (String serializedValue : values) {
-            IValue deserializedValue = valueType.deserialize(serializedValue.replaceAll(ELEMENT_DELIMITER_ESCAPED, ELEMENT_DELIMITER));
+        for (int i = 1; i < split.length; ++i) {
+            if (heterogeneous) {
```
rubensworks, line 65: I think this will not be backwards-compatible with existing lists with category value types.

What we could do instead, is adding a flag to heterogeneous lists. If this flag is not available, deserialization code should be exactly the same as before (with a TODO that is should be removed in 1.13).

In 1.13, we can remove this flag and the backwards-compat code.

src/main/java/org/cyclops/integrateddynamics/core/logicprogrammer/ValueTypeListLPElement.java
```diff
@@ -302,7 +303,11 @@ public void drawGuiContainerBackgroundLayer(int guiLeft, int guiTop, TextureMana
 
         @Override
         public void onChanged() {
-            element.setListValueType(valueTypeSelector.getActiveElement());
+            IValueType newType = valueTypeSelector.getActiveElement();
+            element.setListValueType(newType);
+            if(arrowAdd != null) {
+                arrowAdd.enabled = newType != ValueTypes.CATEGORY_ANY;
```
rubensworks, line 309: What does this do exactly?

rubensworks (changes requested):
My main concern with this PR is backwards-compatibility, but it should be fixable.

See comments inline.

Could you also add some (simple) unit tests on the (de)serialization of heterogeneous lists?

src/main/java/org/cyclops/integrateddynamics/core/logicprogrammer/ValueTypeListLPElement.java
josephcsible: It disables the `+` button when you select to make a List&lt;Any&gt; in the logic programmer, so you can only use it to make empty lists directly. This is because I'm not sure of a good idea to let users build these kinds of lists directly.

rubensworks: It makes sense IMO to allow users to create such lists. But this would require some significant changes in the editor, so I agree that it makes sense to leave this out-of-scope for this PR.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyMaterializedFactory.java
josephcsible: How can any existing serialized lists have category value types? There's currently code in the serialize function that checks if a list has category type, and if so, pretends that it has the type of its first element. (And in the special case of an empty list, this isn't actually backwards incompatible.)

src/main/java/org/cyclops/integrateddynamics/core/logicprogrammer/ValueTypeListLPElement.java
josephcsible: Agreed.

josephcsible:
I'll add tests later today.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeListProxyMaterializedFactory.java
rubensworks: Right, I forgot about that change (this wasn't always so).

I guess it's safe to assume that all lists were created after this change (it was quite a while ago), so no backwards-compat is needed here indeed.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeList.java
```diff
@@ -117,6 +117,10 @@ private ValueList(IValueTypeListProxy<T, V> value) {
             return values.length == 0 ? ValueTypes.LIST.getDefault() : ofList(values[0].getType(), ImmutableList.copyOf(values));
```
josephcsible, line 117: I'm thinking of getting rid of this method in another PR. It's only used by testing code, and all uses can easily be replaced with the new overload I added.

josephcsible:
Tests added.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeList.java
rubensworks: Sounds good!

rubensworks (approved):
Looks good, thanks!

PR #647, opened 2018-11-23
Fix the creative energy battery acting like a regular energy battery when in item form

josephcsible:
This mirrors what TileEnergyBattery already does.


josephcsible:
Side note: should Maintenance have "To Review" like Features does?

rubensworks:
> Side note: should Maintenance have "To Review" like Features does?

Yes! Added it :-)

PR #670, opened 2019-01-18
Fix item stack deserialization

### LINKED ISSUE #669: Display Panel does not show items with stacks bigger than 127
#### Issue type:

- :bug: Bug

____

#### Short description:

A Display Panel only displays stacks that are of size 0 - 127, 128 and higher seems to display nothing ("-128xtileair@0")
This issue also repeats on higher numbers, for example the range 257 - 383 works, further numbers do not.

1. Build a "Display Panel"
2. Make a variable card with an item
3. Make an integer card with value of 128 - 256
4. Make a variable card that uses the first item withSize of the integer card
5. Place all cards except the one made in step 4 into a connected variable store
6. Place card from step 4 into text panel

#### Expected behaviour:

The Display Panel should show the correct item icon and count instead of being blank.

____

#### Versions:

- This mod: 1.12.2-1.0.0-DEV (commit 170bc62dc06bb33657152c31150efc2aca11c5b0)
- Minecraft: 1.12.2
- Forge: 14.23.2.2611

virusek20:
Fixes #669


rubensworks:
Looks great, thanks!

Issue #690, opened 2019-02-08
String Casts


LostOfThought:
#### Issue type:

- :heavy_plus_sign: Feature request

____

#### Short description:

Cast to and from String as other base datatypes such as Int, Float, Long

PR #695, opened 2019-02-12
Initial Implementation for String Parse

LostOfThought:
Functional, but really just a big cast clone....

Ref: #690


rubensworks:
I haven't looked in detail into this yet, but a couple of things I notice atm:

* Parsing only really makes sense from string to something else. So I would remove the handlers for the reverse way. (And refactor so that the input is always a string)
* All ID operators are extensively unit tested, so this would also need unit tests.

LostOfThought:
Renamed to a more generic 'StringConversion'
OP name is dependant on direction (`<type>_as_String`/`parse_<type>`)

I'll get around to unit tests (possab|eventual)ly.

rubensworks:
`<type>_as_String` should not really be needed. You can already convert anything to string using the `name`(or something like that) operator AFAIK.

LostOfThought:
`name` fails for NBT

rubensworks:
> `name` fails for NBT

Sounds like a bug then :-)

LostOfThought:
Not too much more to go.
Let me know if you see any more glaring issues.

LostOfThought:
And yes, spam copy/paste tests lead to several wrong assertion texts.

rubensworks:
I don't see any major issues atm. Will review in detail once the unit tests pass.

PR #745, opened 2019-06-23
Random functions, Absolute

eadgbear:
This PR adds the following functionality:

* `rand` -> Random from zero to any number type
* `rand2` -> Random from any number to any number (result is type of the second)
* `randElem` -> Random element from a list
* `randBool` -> Random boolean
* `abs` -> Absolute value of a number type

It's in need of a pass to make sure it fully conforms with the rest of the library design, but tests are included for each one and they've been tested in-world.


rubensworks:
Thanks! I'll try to look into this asap.

It looks like Travis is failing on the unit tests though.

eadgbear:
@rubensworks I'm a bit confused by the Travis failure, I'm not seeing that exception locally. 

```
org.cyclops.integrateddynamics.core.evaluate.variable.TestGeneralOperators > testRandMixedTypes FAILED
    org.cyclops.integrateddynamics.api.evaluate.variable.IValueCastRegistry$ValueCastException at TestGeneralOperators.java:249
```
I'm not seeing the actual exception message so I can diagnose it. In this test in particular an Integer variable is cast to a Double, which is a valid no-loss operation in Java, but I'm guessing not in the evaluation engine.

rubensworks:
Not sure what is going on there. Might be a Java compiler difference (Travis is using oraclejdk8).

eadgbear:
@rubensworks I just modified the randRange bit to require same types. Worst case users should use the existing casting methods.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -264,6 +266,23 @@ public static void load() {}
                 return ValueTypeInteger.ValueInteger.of(a.getRawValue() - 1);
             }).build());
 
+
+    /**
+     * Number ABS operator with one input and one output
+     */
+    public static final IOperator NUMBER_ABSOLUTE = REGISTRY.register(OperatorBuilder
```
rubensworks, line 273: I would make this part of the `ARITHMETIC OPERATORS`.

Don't use `OperatorBuilder` directly, but create `OperatorBuilders.ARITHMETIC_1` (A variant of `OperatorBuilders.ARITHMETIC_2`).

Make sure to move this declaration to right after `ARITHMETIC_MINIMUM`.

src/test/java/org/cyclops/integrateddynamics/core/evaluate/variable/TestGeneralOperators.java
```diff
@@ -171,4 +172,105 @@ public void testInvalidInputSizeIsNotNullSmall() throws EvaluationException {
...
+
+        IValue res1 = Operators.NUMBER_ABSOLUTE.evaluate(new IVariable[]{i1});
+        assertThat("positive numbers unchanged", ((ValueTypeInteger.ValueInteger) res1).getRawValue(), is(1));
+
+        IValue res2 = Operators.NUMBER_ABSOLUTE.evaluate(new IVariable[]{i2});
+        assertThat("negative numbers changed", ((ValueTypeInteger.ValueInteger) res2).getRawValue(), is(2));
+    }
+
+    @Test
+    public void testAbsoluteDouble() throws EvaluationException {
```
rubensworks, line 191: Can you also add one for Long?

src/test/java/org/cyclops/integrateddynamics/core/evaluate/variable/TestGeneralOperators.java
```diff
@@ -171,4 +172,105 @@ public void testInvalidInputSizeIsNotNullSmall() throws EvaluationException {
         Operators.NULLABLE_ISNOTNULL.evaluate(new IVariable[]{});
     }
 
+    /**
+     * ----------------------------------- ABS -----------------------------------
```
rubensworks, line 176: Can you also add `testInvalidInputSizeIAbsLarge` and `testInvalidInputSizeAbsSmall` tests?
(For consistency with the unit tests for the other operators)

rubensworks (changes requested):
I like the new `abs` operator, for which I just have a couple of minor comments.

However, I have some major concerns with the random-related operators.
The thing is, this mod assumes all operators to be fully deterministic (like most declarative programming languages).
These new operators unfortunately don't maintain this assumption.
Since ID makes use of this assumption for caching operator chains, this will likely break a lot of things.

I agree however that random value generation is useful, which is why a random aspect was added to the extradimensional reader, which is consistent with the deterministic nature of operators. If really needed, more random-related aspects could be added to the extradimensional reader.

rubensworks:
@slyons I'm going to close this PR. I'm open to reusing your implementations in other ways in another PR though. Let's discuss on Discord if you're interested in this.

PR #780, opened 2019-10-16
Change FE to Energy

TomyLobo:
This gets the name and info of the operators for items in line with the corresponding operators for entities.

### DIFF
--- a/src/main/resources/assets/integrateddynamics/lang/en_us.lang
+++ b/src/main/resources/assets/integrateddynamics/lang/en_us.lang
@@ -1071,12 +1071,12 @@ operator.operators.integrateddynamics.entity.entityenergystored.info=The amount 
 operator.operators.integrateddynamics.entity.entityenergycapacity.name=Entity Energy Capacity
 operator.operators.integrateddynamics.entity.entityenergycapacity.info=The energy capacity of this entity.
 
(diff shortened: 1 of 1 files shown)


TomyLobo:
I'd love to sign the CLA, but why does that CLA assistant need my email address? sounds fishy...

rubensworks (changes requested):
I'm fine with this change, but I would consistently replace all FE occurences then, as there seem to be some more usages in the aspect entries.

rubensworks:
> I'd love to sign the CLA, but why does that CLA assistant need my email address? sounds fishy...

@TomyLobo That's just how the CLA assistant handles user identification. I can't merge without it unfortunately.

TomyLobo:
> I'm fine with this change, but I would consistently replace all FE occurences then, as there seem to be some more usages in the aspect entries.

Hmm, I haven't used those, I don't want to change things where I don't know what they do.
Btw, I signed the CLA.

rubensworks:
> Hmm, I haven't used those, I don't want to change things where I don't know what they do.

They are in the same document. A simple `FE` grep will show their location.

TomyLobo:
I know, I found them in the lang file. Doesn't mean I know what they do :)
I think docs should be written by people who know the subject :)

rubensworks:
Closing due to inactivity.

PR #787, opened 2019-11-09
PartNetworkElement: small performance tweak

the-eater:
when profiling my server, 17% of total time (not CPU time, so it might actually be worse) was spend in the `getOrDefault` and `put` functions of the `updateableElementsTicks` map in the `Network` class as (function calls can be seen here [Network.java#L404](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.12/src/main/java/org/cyclops/integrateddynamics/core/network/Network.java#L404))
 
this in part was because of the constant calling to `compareTo` in the `PartNetworkElement` (it's a TreeMap, so it's using sorting for placement yada yada) most of this time (7%~) was spend in `.getClass().getCanonicalName()`, every time this function is called it reconstructs the name without caching making it very slow, so I have replaced those with `.getName()` (parts have a unique name per component) and `.getClass().getName()` which is unique enough for this use case and is cached, so the string building overhead is gone.

currently running my server with this patch applied, the CPU time of `getOrDefault` and `put` has now been reduced to 2.1%

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/network/PartNetworkElement.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/network/PartNetworkElement.java
@@ -206,9 +206,9 @@ public class PartNetworkElement<P extends IPartType<P, S>, S extends IPartState<
 
     @Override
     public int compareTo(INetworkElement o) {
-        if(o instanceof IPartNetworkElement) {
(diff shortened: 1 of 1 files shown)


rubensworks (approved):
Thanks for the PR!

I don't see any potential problems of this change, only benefits, so that's great.

I'll include this in the next ID release.

PR #789, opened 2019-11-18
Filter List for String values before joining

### LINKED ISSUE #783: Server crash exploit
#### Issue type:

- :bug: Bug

____

#### Short description:

Disclosed in a mail to @rubensworks.
I will disclose details here once a fix is released.

1. Build this:
image
2. Throw your equivalent of card #333 into a display panel
3. Throw everthing else into a variable store
    - Expected: Anything but a crash
    - Actual: crash
4. Restart server
    - Expected: Anything but a crash
    - Actual: crash

#### Expected behaviour:

____

#### Versions:

- This mod: IntegratedDynamics-1.12.2-1.1.0.jar
- Minecraft: 1.12.2
- Forge: 14.23.5.2847

#### Log file:

[crash-2019-10-26_17.14.23-server.txt](https://github.com/CyclopsMC/IntegratedDynamics/files/3865296/crash-2019-10-26_17.14.23-server.txt)

TomyLobo:
I also replaced the string joining by a joining collector.

Fixes #783

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -726,8 +726,20 @@ public final class Operators {
                             elements.getRawValue().getValueType(), ValueTypes.STRING);
                     throw new EvaluationException(error.localize());
                 }
+
+                final java.util.Optional<IValueType> firstNonString = Streams.stream(elements.getRawValue())
+                        .map(IValue::getType)
+                        .filter(value -> value != ValueTypes.STRING)
+                        .findFirst();
+
+                if (firstNonString.isPresent()) {
+                    L10NHelpers.UnlocalizedString error = new L10NHelpers.UnlocalizedString(
+                            L10NValues.VALUETYPE_ERROR_INVALIDLISTVALUETYPE,
+                            firstNonString.get(), ValueTypes.STRING);
+                    throw new EvaluationException(error.localize());
+                }
+
                 return ValueTypeString.ValueString.of(Streams.stream(elements.getRawValue())
-                        .filter(value -> value instanceof ValueTypeString.ValueString)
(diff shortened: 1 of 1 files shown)


rubensworks (changes requested):
Not sure silently filtering away the non-string elements is the way to go here. Throwing an eval exception would be the preferred solution when such elements are encountered.

TomyLobo:
@rubensworks I changed it as requested.
And I noticed that correspondTo needs some work.
It's not clear whether it's a necessary or a sufficient criterion for casting everything to string.
Apparently it's only a necessary criterion. That should be documented or the function be given a better name.

Btw, I see a ton of unchecked generics everywhere I look.
That's a ticking time bomb, you basically have no way of telling whether an inocuous assignment somewhere will throw a ClassCastException resulting from an invalid unchecked cast somewhere else.

rubensworks (changes requested):
Thanks for the change, but I feel like this is becoming too complex for what it is.
The code will now be iterating the list two times in regular cases, which is not needed.

What about using a single simple for-loop instead?

rubensworks:
> And I noticed that correspondTo needs some work.

> Btw, I see a ton of unchecked generics everywhere I look.

PRs for improvements are always welcome.

TomyLobo:
> 
> 
> Thanks for the change, but I feel like this is becoming too complex for what it is.
> The code will now be iterating the list two times in regular cases, which is not needed.
> 
> What about using a single simple for-loop instead?

If you want to cut down on CPU cycles, keep in mind that your original code already iterated this twice: Once to create the list and another time to join it.

If you want to cut down code, a more effective approach would be to remove the "correspondsTo" if block, as that no longer catches anything that isn't already caught later.

rubensworks:
I went for a different solution (d0963292bdc3c5af4a2b43d12caefb6cb9afe7a8) that has a lower computational complexity.
Thanks in any case for the PR!

PR #791, opened 2019-12-16
Network diagnostic tool: cleanup and bugfixes

TomyLobo:
The network diagnostics tables sort the tick time by string instead of by actual numerical value.
This results in 10 being sorted lower than 9, for instance.

This PR fixes that.

Before:
image
After:
image

I also got rid of some redundant code and a warning and made some.

~~Full disclosure: I did not test the "Storage Observers" part, since I couldn't get anything to appear there with only Integrated Dynamics and no Integrated Tunnels installed.
Apart from removing that redundant code, the changes to both tables are identical, though, so it *should* work fine. :)~~ I was able to test it using a battery and a furnace.

EDIT:
I made a lot more changes since I wrote that description.
Here's a list of all the changes, from master to current:
- 3 commits - Code quality changes
- Made the hellish `getColumnClass` override a bit less eye-bleeding
- Deduplicate code
- Made the hellish `getColumnClass` override a bit less eye-bleeding
- **Fixed the sorting by tickttime**, as mentioned above
- **Fixed the sorting by position**, too
- **Mark the entire line instead of just certain columns**
- **Use the default selection color instead of dark blue**
- Code quality changes
- **Fix teleporting to full blocks**
- Code quality changes
- Optical changes to the red box around marked components
- Code quality changes
- **Fix red box around marked full-block components**
- **Reduce visual feedback delay when marking components**

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/network/diagnostics/GuiNetworkDiagnostics.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/network/diagnostics/GuiNetworkDiagnostics.java
@@ -269,6 +269,7 @@ public class GuiNetworkDiagnostics extends JFrame {
                                         ObservablePartData partData = getPartDataFromRow(row);
                                         if (partData != null) {
                                             teleportPlayer(e, partData.toPartPos());
+                                            ((DefaultTableModel) target.getModel()).fireTableRowsUpdated(row, row);
                                         }
                                     }
                                 }
@@ -330,6 +331,7 @@ public class GuiNetworkDiagnostics extends JFrame {
                                         ObservableObserverData observerData = getObserverDataFromRow(row);
                                         if (observerData != null) {
                                             teleportPlayer(e, observerData.toPartPos());
+                                            ((DefaultTableModel) target.getModel()).fireTableRowsUpdated(row, row);
                                         }
                                     }
                                 }
@@ -427,6 +429,9 @@ public class GuiNetworkDiagnostics extends JFrame {
     }
 
     private static void teleportPlayer(MouseEvent e, PartPos pos) {
+        if (pos == null) {
+            return;
+        }
         if (e.getButton() == MouseEvent.BUTTON1) {
             if (NetworkDiagnosticsPartOverlayRenderer.getInstance().hasPartPos(pos)) {
                 NetworkDiagnosticsPartOverlayRenderer.getInstance().removePos(pos);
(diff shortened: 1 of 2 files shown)


TomyLobo:
I can fix the sorting of the position column in the same way if you want.
But that won't be in the order it is now. Here's a simplified version of Vec3i#compareTo:
```java
    public int compareTo(Vec3i other) {
        if (this.y != other.y) {
            return this.y - other.y;
        } 
        if (this.z != other.z) {
            return this.z - other.z;
        } 
        return this.x - other.x;
    }
```
Here's the result:
image

TomyLobo:
Please hold off merging this for a bit more. I'm refactoring it a bit, while editing old commits and stuff

TomyLobo:
Lots of changes, I updated the description with a list of them, with the important ones in bold.
The list is ordered like the Git history.

I'm done now, I think, so you can merge now if you want :)

rubensworks:
Thanks! I'll try to look into this later this week.

src/main/java/org/cyclops/integrateddynamics/core/network/diagnostics/GuiNetworkDiagnostics.java
```diff
@@ -263,67 +285,83 @@ public void mouseClicked(MouseEvent e) {
...
                             });
 
+                            for (JTable table : new JTable[] { tableParts, tableObservers }) {
+                                table.getColumnModel().removeColumn(table.getColumn("_id"));
+                                table.setAutoCreateRowSorter(true);
+
+                                table.setDefaultRenderer(Double.class, new DefaultTableCellRenderer() {
+                                    @Override
+                                    public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int column) {
+                                        //noinspection MalformedFormatString
```
rubensworks: Just casting to `Double` makes this unnecessary.

rubensworks (changes requested):
I'm quite happy with these changes. It definitely improves the code a lot (it was quite horrible...).

It looks like this PR makes the ticktime column align text to the left. It would be great if you could change this to align to the right like before, to make it better readable.

I'll probably squash all of these commits before merging, as commits are a bit all over the place, and don't fit well with the current commit structure of ID. (But feel free to rewrite history yourself if you don't want me to squash)

src/main/java/org/cyclops/integrateddynamics/core/network/diagnostics/GuiNetworkDiagnostics.java
TomyLobo: But then IntelliJ will complain about the superfluous cast :)

So, since it complains in both cases, I picked the alternative with less casts.

TomyLobo:
> 
> 
> I'm quite happy with these changes. It definitely improves the code a lot (it was quite horrible...).
> 
> It looks like this PR makes the ticktime column align text to the left. It would be great if you could change this to align to the right like before, to make it better readable.
> 
> I'll probably squash all of these commits before merging, as commits are a bit all over the place, and don't fit well with the current commit structure of ID. (But feel free to rewrite history yourself if you don't want me to squash)

I'll look into alignment tomorrow.
My advice would be to keep these commits intact (or at least squash them to more than one commit) since that makes Git archeology so much easier if every commit only changes one thing.
If you want a big picture of one particular commit, GitHub will show a link to the PR that allows just that.

TomyLobo:
Alright, there's the alignment and while looking for it, I figured out a better way to override the renderer and I got rid of the //noinspection.

rubensworks (approved):
> My advice would be to keep these commits intact (or at least squash them to more than one commit)

Feel free to squash in to larger commits, I don't have the bandwidth to go through these myself unfortunately. Otherwise I'll just squash into one commit tomorrow.

rubensworks:
Thanks!
I'll include this in the next release.

PR #857, opened 2020-05-09
Add config options for power consumption of individual network components

### LINKED ISSUE #856: More Configurability of Network Energy Consumption
#### Issue type:

____

#### Short description:

I'm planning on assembling a personal modpack in which ID network power consumption is enabled, but I noticed that the relative energy consumption values for each network part are hard-coded. For example, I'd like to make storage terminals rather energy-hungry, but I can't do so without making variable displays require an equal amount of energy, and variable stores would require twice as much energy. It would be nice if I could manually set the relative energy consumption values for each part in the configs.

Trainerredstone7:
Resolves #856.


src/main/java/org/cyclops/integrateddynamics/part/PartTypePanelDisplay.java
```diff
@@ -32,7 +33,7 @@ protected Block createBlock(BlockConfig blockConfig) {
 
     @Override
     public int getConsumptionRate(State state) {
-        return state.hasVariable() ? 2 : 1;
+        return GeneralConfig.panelDisplayBaseConsumption;
```
rubensworks: Could you add `panelDisplayBaseConsumptionEnabled` and `panelDisplayBaseConsumptionDisabled` here instead?

rubensworks (changes requested):
Looks good!

Having separate fields for every part is a bit unfortunate, but it's the best we can do with the current config system I guess. I should be able to improve this in 1.15 though.

src/main/java/org/cyclops/integrateddynamics/part/PartTypePanelDisplay.java
Trainerredstone7: I've split the display panel config option into separate options for enabled and disabled states.

Trainerredstone7:
I'll work on making similar pull requests to add power consumption config options to Integrated Tunnels, Crafting, and Terminals.

rubensworks:
Great, thanks!

PR #873, opened 2020-05-28
Fix NPE caused by not checking .isPresent

SCLeoX:
The crash can be reproduced by using any of the three NBT path related operator on a null NBT variable.


rubensworks:
Thanks!

Issue #180, opened 2017-05-21
Facade blocks are not oriented correctly.


VampireValik:
- Minecraft: 1.11.2
- forge1.11.2-13.20.0.2299
- IntegratedDynamics-1.11.2-0.7.1
- Lots of other mods, all up-to-date.

I've noticed that some facade blocks are not oriented correctly. I first noticed the problem with Environmental Tech Alabaster Bricks. See the screenshot below. On the right is a real Alabaster Brick and on the left is the facade version with a cable inside. I believe that the face I'm looking at in the screenshot should be the top face, not the northern face.
![2017-05-21_18 48 48](https://cloud.githubusercontent.com/assets/28024586/26288045/7f8c2448-3e56-11e7-80d6-5e730dd46ea3.png)

ouroborus:
They don't take on lighting/shadowing properly either.

Here's a screenshot of facades of vanilla bricks:
![2018-01-09_23 04 07](https://user-images.githubusercontent.com/392047/34759727-ba03c106-f591-11e7-96b1-6739897191ac.png)

JHawkley:
Considering this is a two year old issue and still has no movement, perhaps dropping the use of mod-specialized facades and instead having the logic cables and parts support [MCMultiPart](https://github.com/amadornes/MCMultiPart) might be a good alternative.

This is the approach taken by [Mekanism](https://github.com/mekanism/mekanism) for hiding its pipes, so it wouldn't be an especially unfamiliar concept to the player.

People could use something like [Chisels & Bits](https://github.com/AlgorithmX2/Chisels-and-Bits) to craft their own facades around their Integrated Dynamics contraptions.  Integrated Dynamics could also provide a tool to the player that would allow them to easily place fitting facades created from bits, if Chisels & Bits is installed.

Since facade rendering would no longer be handled by Integrated Dynamics, this would likely solve both the texture and lighting issues and vastly increase options for decoration available to the player.

rubensworks:
Fixing support for MCMP is [planned](https://github.com/CyclopsMC/IntegratedDynamics/issues/157), but not trivial. Fixing this lightning issue will probably be a lot easier.

I will never put a hard dependency on third-party mods, so MCMP will always remain optional, which is why I still want to have facades in this mod.

nitelite:
Just wanted to mention that I also bumped into some strange lighting with facades recently in the MC Eternal pack. Probably same issue as mentioned above :)

PR #956, opened 2021-01-02
Fix facades rendering, #180

ivxvm:
This is my attempt to fix facades. Currently I've been able to fix orientation and lightning and the issue with blocks having multiple textures (previous implementation only ever use single texture). Overall I think this covers most cases. There are still some minor rotation issues with blocks which have weird face rotations like terracota, which I was not able to figure out (tried a lot of stuff with no success). I'm not sure if this should be considered as closing #180 or not, maybe the remaining weird-uv-blocks can be dealt with in separate issue.

Before:

After:

What I was not able to fix:


rubensworks:
Oh, this looks great, I'll have a look at it in a bit.

rubensworks:
Thanks a lot for this!

The cases you were not able to fix are due to blocks with non-default block states, which would require some more invasive changes to fix. But your PR already makes things a whole lot better :-)

I'll make a new release once I find some time to resolve the other open 1.12 issues.

Issue #982, opened 2021-01-27
"NullPointerException" Crash when trying to determine the mod name of an block a entity is looking at, when the entity isn't present.


Cerbrus:
#### Issue type:

- :bug: Bug 

____

#### Short description:

I'm walking through the "Advanced Operations" tutorial of the mod. I figured I'd try and combine what I learned so far, and wanted to display the mod name of the item I'm currently looking at. This worked, but the game crashed when I stepped away from the entity detector. The game crashes now.

#### Steps to reproduce the problem:

- Get an entity detector
- Grab a single entity from that
- Proxy that variable into a new variable.
- Create a "target block" variable from that proxy.
- Create a "Mod" variable from that target block.
- Stuff the variables into your network, the display should show the mod name of the block you're looking at.
- Walk away from the entity detector... Boom! The game crashes.

#### Expected behaviour:

I'd expect the error (null reference perhaps) to be handled and the display to show nothing.

____

#### Versions:

- This mod:  
IntegratedDynamics-1.16.4-1.4.0
- Minecraft:  
1.16.4
- Forge:  
forge-1.16.5-geckolib-3.0.4a

I'm on version 1.5.1 of the FTB Direwolf 20 1.16 modpack.

#### Log file:

[crash-2021-01-27_23.00.28-server.txt](https://github.com/CyclopsMC/IntegratedDynamics/files/5883400/crash-2021-01-27_23.00.28-server.txt)

I could provide a savegame, if that helps.

rubensworks:
Thanks for reporting!

Cerbrus:
Well that's quick!

Is there anything I can do that'd make my world load again (preferably without disabling the mod, I wouldn't wanna loose what I built)

Here's the save, in case it helps:

https://drive.google.com/file/d/1RnKH_TNhYU76pjoLn3I10fvhO1t05_Ch/view?usp=sharing
(It's ~650MB)

PR #983, opened 2021-01-28
* Fix a NullReferenceException (#982)

Cerbrus:
This fixes the issue I have with #982.

I'm not sure if this is the correct approach, but it at least allowed me to load back into the game :-)


rubensworks:
Thanks for the PR. Unfortunately, this is not the correct fix though. Value propagators should return non-null values. So if there's a null, there's something going wrong in an operator.

Cerbrus:
No problem, I've seen how you fixed the issue, that makes a lot of sense :-)

PR #1049, opened 2021-06-01
Change the block reader's update type to make it to work with IntegratedProxy

shBLOCK:
Change the block reader's update type to make it to work with IntegratedProxy.


rubensworks:
Not sure if this is the best possible solution, as this will have an impact on performance, especially this in the majority of cases, the integrated proxy block will not be used.
Is there any other way around this? E.g., by letting the proxy manually invalidate aspects/variables when needed?

Could you target this change to the 1.16 branch? The 1.12 version is not being maintained anymore.

shBLOCK:
It seems like the AspectUpdateType.NEVER is only used here, is this really a big performance issue?
I just want the 1.12 version to work with it, but if this really need a big effort to change, it's not a big deal.

rubensworks:
> is this really a big performance issue?

It can be, yes. For chunks that update regularly, these aspects will have to be re-calculated every time, instead of only once.
So manual invalidation seems like a better approach to me.

shBLOCK:
Ok, I'll try to do that.

Issue #1152, opened 2022-04-09
Performance Optimization Discussion


Geo25rey:
#### Issue type:

- :snail: Performance issue 

____

#### Short description:

Continuing conversation from #1147. When Having large networks, performance degrades. I found some places where optimizations can be made using [this performance profile](https://spark.lucko.me/fsJzpQ3vc5)

#### Potential Optimizations

The below were brought to my attention from the profile linked above since they were taking a significant amount of time to process/execute.

1. [org.cyclops.integrateddynamics.core.part.aspect.build.AspectBuilder lines 418 and 420](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.16/src/main/java/org/cyclops/integrateddynamics/core/part/aspect/build/AspectBuilder.java#L418) can be run in parallel since there doesn't seem to be any order requirements in each `onUpdate` method.

2. [org.cyclops.integrateddynamics.core.network.Network line 441](https://github.com/CyclopsMC/IntegratedDynamics/blob/9998765132a2fd55a243118ac1a4c00b8dc230b9/src/main/java/org/cyclops/integrateddynamics/core/network/Network.java#L441) same recommendation as 1.

3. [org.cyclops.integrateddynamics.core.network.IngredientChannelAdapterWrapperSlotted lines 52 and 55](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.16/src/main/java/org/cyclops/integrateddynamics/core/network/IngredientChannelAdapterWrapperSlotted.java#L52L55) Setting the position state to disabled then re-enabled in the same context seems unnecessary since those changes can only happen on the main thread in series anyway.

4. [org.cyclops.integrateddynamics.core.network.IngredientChannelAdapterWrapperSlotted lines 69 and 72](https://github.com/CyclopsMC/IntegratedDynamics/blob/9998765132a2fd55a243118ac1a4c00b8dc230b9/src/main/java/org/cyclops/integrateddynamics/core/network/IngredientChannelAdapterWrapperSlotted.java#L69L72) same recommendation as 3.

rubensworks:
Thanks for reporting!

rubensworks:
Thanks for all the ideas @Geo25rey, I'll reply to each of them below:

[...]

PR #1153, opened 2022-04-09
[1.16] performance: remove disabling of inventory position when getting ingredient channel slots

Geo25rey:
See discussion: #1152

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/network/IngredientChannelAdapterWrapperSlotted.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/network/IngredientChannelAdapterWrapperSlotted.java
@@ -49,10 +49,8 @@ public class IngredientChannelAdapterWrapperSlotted<T, M> implements IIngredient
             if (!pos.getPos().isLoaded() || network.isPositionDisabled(pos)) {
                 continue;
             }
-            network.disablePosition(pos);
             IIngredientComponentStorage<T, M> storage = network.getPositionedStorage(pos);
             slots = Helpers.addSafe(slots, getIngredientComponentStorageSize(storage));
-            network.enablePosition(pos);
         }
 
         return slots;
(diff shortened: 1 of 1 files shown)


Geo25rey:
@rubensworks can you run the CI? I can't get the build config to work locally

rubensworks (changes requested):
I'm afraid these method calls have to remain in-place, because they are needed to handle the case of infinite loops for self-recursive networks (e.g. IT item interfaces are connected to each other directly).
Without them, any call to `getSlots()` on such network will crash the server with a stackoverflow exception.

Geo25rey:
> to handle the case of infinite loops for self-recursive networks

Sounds like this should be a compiled state on block and slot updates. Compiling the state of the network from scratch on every tick seems very wasteful to me. Not sure the best way to go about this since this would likely be a big refactor on your part.

Geo25rey:
> Without them, any call to `getSlots()` on such network will crash the server with a stackoverflow exception.

On the other hand, this could also be a UX issue rather than a technical one. There could be a check when a user updates the network to see if there is any infinite loops possible and prompt the user that doing whatever update they are making will break their network and/or not actually not actually complete the action that the user just made.

rubensworks:
> Sounds like this should be a compiled state on block and slot updates. Compiling the state of the network from scratch on every tick seems very wasteful to me. Not sure the best way to go about this since this would likely be a big refactor on your part.

The state _is_ precompiled. Recalculation only happens when placing new cables.

rubensworks:
> On the other hand, this could also be a UX issue rather than a technical one. There could be a check when a user updates the network to see if there is any infinite loops possible and prompt the user that doing whatever update they are making will break their network and/or not actually not actually complete the action that the user just made.

Yes, something like this may indeed be better. But this might not be that easy to achieve though, since loops could also be achieved with hops through other networks (even from other mods). But I'm definitely open to improvements in this regard.

Related to https://github.com/CyclopsMC/IntegratedTunnels/issues/69

rubensworks:
Closing due to inactivity.

PR #1165, opened 2022-05-11
Korean Localization

gjeodnd12165:
It's difficult to localize the error message with format(%s). Can I set the order of formatting characters like %1s, %2s, ... ?


rubensworks:
Thanks!

> Can I set the order of formatting characters like %1s, %2s, ... ?

No, that's not possible unfortunately.

PR #1171, opened 2022-05-23
Merge Network updatable element set+map

octylFractal:
This reduces the need to do two tree lookups every tick to adjust the tick times, and saves memory by only having one map and using an optimized fastutil map.

I tested these changes on my server, and it saved me 5ms per tick on a large ID network, so it's a pretty significant impact.


rubensworks (approved):
This is really neat, thanks for this @octylFractal!

I'll make a new release with this improvement soon.

Issue #1217, opened 2022-09-26
Add shorted number formatting


nskobelevs:
#### Issue type:

- :heavy_plus_sign: Feature request 

____

#### Short description:

A common use of a Display Panel is to display a number and in some cases (e.g. energy count) this numbers can get too large for the panel, impacting readability.

My suggestion is to add an operator to format numbers in a compact way.
Perhaps a second input specifying the number of decimal digits.

|Input|Output|
|:-:|:-:|
|1,254|1.2K|
|5,386,452|5.3M|
|3,385,215,385|3.3B|

Java implements this natively since Java 12 which should make this a relatively easy feature to implement

```java
import java.text.NumberFormat;
import java.util.Locale;
 
public class CompactNumberFormatExample {
 
    public static void main(String args[]) {
 
        NumberFormat nf = NumberFormat.getCompactNumberInstance(
                Locale.US, NumberFormat.Style.SHORT);
                
        nf.setMinimumFractionDigits(3);
        
        System.out.println("Result: " + nf.format(1254));
        System.out.println("Result: " + nf.format(5386452));
        System.out.println("Result: " + nf.format(3385215385L));
    }
}
```

```
Result: 1.254K
Result: 5.386M
Result: 3.385B
```

rubensworks:
Thanks for the suggestion!

shadow7412:
I've raised a PR for this at https://github.com/CyclopsMC/IntegratedDynamics/pull/1221

PR #1221, opened 2022-10-08
#1217 Add 'fuzzy' operator, for shortening numbers to make them useful on displays.

shadow7412:
I've seen this feature request thrown around discord a bit, and figured it'd be a nice one to introduce myself to this codebase.

I (probably) haven't discussed with with the lead developers, as advised by the readme, but seeing as this wasn't a large task I figure we can start this decision here :smile: 

What languages are people expected to support out of the box, as I personally am under qualified to give a meaningful translation to non-english (and I wouldn't presume to just use google translate and hope for the best, unless that's your advice...)

image

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -823,6 +823,15 @@ public final class Operators {
             ).build());
 
     /**
+     * Accepts a number, and returns a string roughly representing that number
+     */
+    public static final IOperator NUMBER_FUZZY = REGISTRY.register(OperatorBuilders.NUMBER_1_LONG
+            .inputType(ValueTypes.CATEGORY_NUMBER).output(ValueTypes.STRING).symbolOperator("fuzzy")
+            .function(
+                variables -> ValueTypes.CATEGORY_NUMBER.fuzzy(variables.getVariables()[0])
+            ).build());
+
+    /**
      * ----------------------------------- NULLABLE OPERATORS -----------------------------------
      */
 
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -90,7 +90,6 @@ import org.cyclops.integrateddynamics.core.helper.NbtHelpers;
 import org.cyclops.integrateddynamics.core.ingredient.ExtendedIngredientsList;
 import org.cyclops.integrateddynamics.core.ingredient.ExtendedIngredientsSingle;
 
-import java.awt.*;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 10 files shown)


src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -822,6 +822,15 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
                 variables -> ValueTypes.CATEGORY_NUMBER.floor(variables.getVariables()[0])
             ).build());
 
+    /**
+     * Accepts a number, and returns a string roughly representing that number
+     */
+    public static final IOperator NUMBER_FUZZY = REGISTRY.register(OperatorBuilders.NUMBER_1_LONG
```
rubensworks: Not sure about the name. Could we use something like `compact` instead?

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeDouble.java
```diff
@@ -125,6 +128,13 @@ public ValueTypeInteger.ValueInteger floor(ValueDouble a) {
         return ValueTypeInteger.ValueInteger.of((int) Math.floor(a.getRawValue()));
     }
 
+    @Override
+    public ValueTypeString.ValueString fuzzy(ValueDouble a) {
+        NumberFormat nf = NumberFormat.getCompactNumberInstance(Locale.US, NumberFormat.Style.SHORT);
```
rubensworks: The compacting is very opinionated here.
I suggest moving _all_ number format options to the config file (see `GeneralConfig`), so things like the style, max/min fract digits, max/min int digits.

rubensworks (changes requested):
Hi @shadow7412, thanks for the contribution!
The feature itself looks definitely like something we need.
Some minor comments in the review below.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeDouble.java
```diff
@@ -125,6 +128,13 @@ public ValueTypeInteger.ValueInteger floor(ValueDouble a) {
...
+        nf.setMaximumFractionDigits(2);
+        return ValueTypeString.ValueString.of(nf.format(a.getRawValue()));
```
rubensworks, line 142: Also, since the implementation seems identical across all number types, perhaps we can move this to a `default` interface method.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeDouble.java
shadow7412: That's a suspiciously convenient configuration system...

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeDouble.java
shadow7412: I might need some help with that one - when I attempted to move it into `IValueTypenumber.java`, `getRawValue` isn't defined...

I've made the config change in the 3 places in case you decide to relax this suggestion (though if there is some way to do this that I'm missing, I totally agree with it)

rubensworks:
Thanks, looks great! Will prepare a new release in a bit.

Issue #1141, opened 2022-03-05
Intersection of lists A and B


DrSigma4164:
#### Issue type:

- :heavy_plus_sign: Feature request

____

#### Short description:
Add an operator that will select only the elements common to the two lists. That is, the Intersection of lists A and B.
Or if it will be too difficult to implement then the operator who simply tells whether there is at least one mpilny element in both lists.

rubensworks:
Thanks for the suggestion!

shadow7412:
I've proposed an implementation of this at #1224.

shadow7412:
This has been completed (cc: @rubensworks)

PR #1224, opened 2022-10-09
#1141 Added intersection operator.

shadow7412:
image

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -100,6 +100,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Collection of available operators.
@@ -1180,6 +1181,24 @@ public final class Operators {
                     throw new EvaluationException(Component.translatable(L10NValues.OPERATOR_ERROR_SLICE_INDEXNEGATIVE));
                 }
                 return ValueTypeList.ValueList.ofFactory(new ValueTypeListProxySlice<>(list, from.getRawValue(), to.getRawValue()));
+            }).build());
+
+    public static final IOperator LIST_INTERSECTION = REGISTRY.register(OperatorBuilders.LIST
+            .inputTypes(ValueTypes.LIST, ValueTypes.LIST)
+            .renderPattern(IConfigRenderPattern.PREFIX_2_LONG).output(ValueTypes.LIST)
+            .symbol("∩").operatorName("intersection")
+            .function(variables -> {
+                ValueTypeList.ValueList list1 = variables.getValue(0, ValueTypes.LIST);
+                IValueTypeListProxy<IValueType<IValue>, IValue> rawList1 = list1.getRawValue();
+                ValueTypeList.ValueList list2 = variables.getValue(1, ValueTypes.LIST);
+                IValueTypeListProxy<IValueType<IValue>, IValue> rawList2 = list2.getRawValue();
+                Stream<IValue> result = new ArrayList<>(Sets.newLinkedHashSet(rawList1)).stream().filter(value1 -> {
+                    for (IValue value2: rawList2) {
+                        if (value1.equals(value2)) return true;
+                    }
+                    return false;
+                });
+                return ValueTypeList.ValueList.ofList(rawList1.getValueType(), result.toList());
             }).build());
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1182,6 +1183,24 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
...
+    public static final IOperator LIST_INTERSECTION = REGISTRY.register(OperatorBuilders.LIST
+            .inputTypes(ValueTypes.LIST, ValueTypes.LIST)
+            .renderPattern(IConfigRenderPattern.PREFIX_2_LONG).output(ValueTypes.LIST)
+            .symbol("∩").operatorName("intersection")
+            .function(variables -> {
+                ValueTypeList.ValueList list1 = variables.getValue(0, ValueTypes.LIST);
+                IValueTypeListProxy<IValueType<IValue>, IValue> rawList1 = list1.getRawValue();
+                ValueTypeList.ValueList list2 = variables.getValue(1, ValueTypes.LIST);
+                IValueTypeListProxy<IValueType<IValue>, IValue> rawList2 = list2.getRawValue();
+                Stream<IValue> result = new ArrayList<>(Sets.newLinkedHashSet(rawList1)).stream().filter(value1 -> {
```
rubensworks: Why is a set used here? We probably don't need it I guess.

rubensworks (changes requested):

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
shadow7412: Actually good point. I've refactored this to use set methods, which are bound to be more efficient - with a bonus of fixing an issue where and item being present...

That said, it also raises the question of what the expected behaviour would be when multiple item are present in a given list.
image
Should that show a single 0 (ie, there are 0s in both lists) or two 0s (because there is a minimum of two 0s in both lists)?

I've gone with option A (there are 0s in both lists, therefore have a 0) - but what do you think?

rubensworks: The problem is that `IValue` does not implement `hashCode` (AFAIK), so the performance benefit will be minimal.

> That said, it also raises the question of what the expected behaviour would be when multiple item are present in a given list.

Perhaps it's better to leave that behaviour undefined. If needed, the user can still use `uniq`.

shadow7412: Does that mean you want me to pull back that last commit, and get rid of the sets?

rubensworks: I guess we can keep the new version. But definitely test it as well with more complex datatypes (like recipes), to make sure we don't run into issues due to the lacking `hashCode`.

shadow7412: I just tried recipes - seems to work. Making a list of recipes is a pain though :P

There may be performance gains to be had (in this, and perhaps other operators) but it doesn't seem like it's *required* at this stage.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1182,6 +1174,20 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
...
+            .inputTypes(ValueTypes.LIST, ValueTypes.LIST)
+            .renderPattern(IConfigRenderPattern.PREFIX_2_LONG).output(ValueTypes.LIST)
+            .symbol("∩").operatorName("intersection")
+            .function(variables -> {
+                IValueTypeListProxy<IValueType<IValue>, IValue> rawList1 = variables.getValue(0, ValueTypes.LIST).getRawValue();
+                IValueTypeListProxy<IValueType<IValue>, IValue> rawList2 = variables.getValue(1, ValueTypes.LIST).getRawValue();
+
+                LinkedHashSet<IValue> result = Sets.newLinkedHashSet(rawList1);
+                result.retainAll(Sets.newLinkedHashSet(rawList2));
+
```
rubensworks, line 1192: Just one more thing I just thought of: we need to error if one of the lists is infinite.
Otherwise, the server may crash.

Similar to this: https://github.com/CyclopsMC/IntegratedDynamics/blob/daac9ddde2364c0068807afad36ca86ac0b5c9ee/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java#L979-L982

shadow7412: I'm a little stumped by this;
image
I can't tell what I'm doing differently to your example which doesn't raise this error...

shadow7412: Never mind, worked it out. You're making an IFunction instead of using the lambda directly.

rubensworks:
Excellent, thanks!

PR #1285, opened 2023-07-15
Give `DOUBLE` higher precedence than `LONG` for number operations

### LINKED ISSUE #1284: Type cast precedence for arithmetic operators
#### Issue type:

- :question: Question 

____

#### Question:

The Arithmetic Addition operator (and the other arithmetic operators) supports adding numbers of different types, e.g. `Int`, `Long`, and `Double`. [.../core/evaluate/variable/ValueTypeCategoryNumber.java](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.20/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java) looks to be the relevant bit in the code. The process seems to be to cast all numerical inputs to the same type ([L79-L80](https://github.com/CyclopsMC/IntegratedDynamics/blob/a315e4d1d4fec1a8c0978ce1c33dfcc6c8bc531a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java#L79-L80)) as that of the type with the highest index according to the `ELEMENTS` list ([L42-L57](https://github.com/CyclopsMC/IntegratedDynamics/blob/a315e4d1d4fec1a8c0978ce1c33dfcc6c8bc531a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java#L42-L57)).

The current `ELEMENTS` list is `{ValueTypes.INTEGER, ValueTypes.DOUBLE, ValueTypes.LONG}`, which means that any arithmetic operators that involve longs will cause every other value to be cast to a long - including doubles. This leads to behaviour such as `1.6` (double) + `2` (long) = `3` (long) - truncating the double as it gets cast, and the final result being a long. The screenshot below shows the result of both `1.6 + 2` and `2 + 1.6`.
image

Using `{ValueTypes.INTEGER, ValueTypes.LONG, ValueTypes.DOUBLE}` (i.e. double having higher precedence than long) would both avoid implicitly truncating floating point numbers, and have the result of operations with floating point numbers also be a floating point number - both of which might be expected, based on how other programming languages work. Is it possible to use this ordering over the current one?

met4000:
Closes #1284

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java
@@ -23,7 +23,7 @@ import java.util.Map;
  */
 public class ValueTypeCategoryNumber extends ValueTypeCategoryBase<IValue> implements IValueTypeNamed<IValue> {
 
-    private static final IValueTypeNumber[] ELEMENTS = new IValueTypeNumber[]{ValueTypes.INTEGER, ValueTypes.DOUBLE, ValueTypes.LONG};
+    private static final IValueTypeNumber[] ELEMENTS = new IValueTypeNumber[]{ValueTypes.INTEGER, ValueTypes.LONG, ValueTypes.DOUBLE};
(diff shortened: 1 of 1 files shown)


rubensworks:
Thanks @met4000! :)

PR #1291, opened 2023-08-04
Integer operation generalisations

met4000:
Generalises the `increment`, `decrement`, and `modulus` operations from `Integer` operations to `Number` operations, allowing them to be used by other number types. Uses the built-in java operator `%` for the implementation of the `modulus` operation of each number type.


src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java
```diff
@@ -147,6 +147,31 @@ public IValue min(IVariable a, IVariable b) throws EvaluationException {
...
+        return type.decrement(castValue(type, a.getValue()));
+    }
+
+    public IValue modulus(IVariable a, IVariable b) throws EvaluationException {
+        IValueTypeNumber type = getLowestType(getType(a), getType(b));
+        IValue bValue = castValue(type, b.getValue());
+        if (type.isZero(bValue)) { // You can not divide by zero
+            throw new EvaluationException(new TranslatableComponent(L10NValues.OPERATOR_ERROR_DIVIDEBYZERO));
+        } else if (type.isZero(bValue)) { // If b is neutral element for division
+            return type.getDefault(); // ? maybe each type should have a 'zero' value defined, rather than just using the default value here
```
met4000: The default value of all current number types is a zero value so I used that here, but I don't know if that's the strictly intended behaviour - should a `getZero` method be added to the number types and used here?

met4000:
In addition to the titular changes, there were a few comments and strings in the tests that seemed to have the wrong value in them - I assume either from copy-pasting, or being in a section that has been renamed without them also being updated. I've modified them here, but I can revert if that's incorrect.

src/main/java/org/cyclops/integrateddynamics/core/evaluate/variable/ValueTypeCategoryNumber.java
rubensworks: Yeah, zero is safe to assume for the default value. Perhaps we may want to add it to the javadoc even.

rubensworks (changes requested):
While I certainly like this change, it contains breaking changes for existing worlds based on how it's implemented now.

So either we wait with merging this until the next major MC update, or we make it backwards-compatible.
One option to make it backwards-compatible would be to keep the integer variants of the operators, but annotate them in the source code with something like `// TODO: remove in next major update`

rubensworks:
Awesome, nice work @met4000!
Will look into preparing new releases later today.

PR #1294, opened 2023-08-09
#1292 cast tests

### LINKED ISSUE #1292: Arithmetic Cast Precedence Tests
#### Issue type:

- :question: Question 

____

#### Question:

In #1285 (implementing #1284), I forgot to add any tests for that ordering. In particular checking Integer -> Long, Integer -> 
Double, and Long -> Double, but maybe also Integer -> Integer etc. (that final type could probably be done by iterating over the list of all the types). The tests would flag if this ordering gets inadvertently changed, and the final test type would also catch if something breaks in e.g. `getLowestType` and suddenly all arithmetic operators cast to only one output type.

Would you accept a PR adding these tests - and if so, would any of those specific tests be redundant?

met4000:
Resolves #1292; adds tests for #1285


rubensworks:
Thanks @met4000! :-)

PR #1389, opened 2024-08-31
Combine cable voxel shape components in getShape rather than getCollisionShape

embeddedt:
This PR moves `BlockCable`'s conversion from `VoxelShapeComponents` to a vanilla `VoxelShape` into the more generic `getShape` function, rather than `getCollisionShape`. This makes the conversion apply to the block's occlusion shape as well and not just the collision shape.

This is needed to fix a compatibility issue with Moonrise which cannot handle the custom `VoxelShapeComponents` implementation. I suspect it might also fix some as-of-yet-unknown issue with cable occlusion, as the occlusion logic uses VoxelShape operations just like entity collisions.


rubensworks:
Thanks @embeddedt! I will release this soon.

Issue #1424, opened 2024-11-24
Recipe for resetting facade texture


AlisonHuang777:
#### Issue type:

- :heavy_plus_sign: Feature request 

____

#### Short description:

As the title described. For now after a facade has crafted with another block and turns into its texture, there's no way for the facade to return to its original no-texture form. This feature will be useful for storing disassembled facades. Such recipe might be as simple as putting a single facade into the crafting grid.

rubensworks:
Thanks for the suggestion!

PR #1441, opened 2024-12-19
Master 1.19 lts

kirjorjos:
Closes issue #1424

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
         with:
           submodules: true
       - name: 'Cache'
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: |
             ~/.m2
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-mod_version=1.24.1
+mod_version=1.24.2
 api_version=0.2.0
 minecraft_version=1.19.2
 forge_version=43.0.8
--- a/resources/changelog/1.19.2-1.24.2.txt
+++ b/resources/changelog/1.19.2-1.24.2.txt
@@ -1,0 +1,10 @@
+As always, don't forget to backup your world before updating!
+Requires CyclopsCore version 1.18.2 or higher.
+
+Additions:
+* Add error operator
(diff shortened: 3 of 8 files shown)


rubensworks (changes requested):
Some thoughts on this PR:

- The nbt-clear approach is very straightforward, and seems like a good option. However, this will cause the original block the facade was bound to, to be lost. We could introduce a custom recipe handler that returns the original block as container item (similar to recipes returning empty buckets). The main problem of that approach, is that clearing facades in bulk won't be possible anymore due to the container item being placed in the crafting grid. So bulk crafting would become very painful in that case. Not sure that we best approach here is, as both have their pros and cons.
- This PR seems to contain two changes (recipe + operator change). Could these two be split into separate PRs? This keeps the commit history (and changelog) a bit cleaner.
- I noticed that you are re-merging your previously merged master branch, which will create a weird commit history. It's better to fetch from upstream and checkout the remote master branch, create a new feature branch from that, and PRing that feature branch. Then, only your new commits will be added to the PR.

kirjorjos:
I can definitely make sure to fix my branch setup and make a new branch for each PR, as for obtaining the block back, we might be able to allow the user to squeeze the block out of the facade and keep losing the block as a convenience recipe?  Failing that, I'm fine implementing either method you described above.

rubensworks:
> we might be able to allow the user to squeeze the block out of the facade and keep losing the block as a convenience recipe?

So if I understand correctly, you suggest keeping the current approach (losing the block)? If so, I can live with that :-)

kirjorjos:
I was suggesting 2 recipes, a convenience that is lossy and what we have now, and a squeezer recipe that isn't

rubensworks:
Sure, two recipes can work. But how shall we distinguish them so they have different inputs?

kirjorjos:
I was thinking the convenience one being in the normal 2x2 crafting grid, and the one that returns the block being a mechanical squeezer recipe, and if it supports multiple outputs (I don't remember without looking), a non-mechanical squeezer recipe as well.

rubensworks:
Ah yes, that's a good idea!

PR #1445, opened 2024-12-23
Clear facade recipes

### LINKED ISSUE #1424: Recipe for resetting facade texture

kirjorjos:
Closes #1424 by adding a normal crafting recipe for lossy clearing and a recipe in both squeezers for non-lossy clearing


rubensworks (changes requested):
I understand the need for a custom recipe serializer, but I'm pretty sure the custom recipe type is not necessary. Could you try using the existing recipe type? I think this will simplify this code by a lot.

kirjorjos:
I could modify the 2 existing recipe types to directly call the calculator method, but the reason I did it with an additional recipe type is to make it clear in the JSON recipe files that it's a special type.  If you think that just putting it in the special folder twice won't cause too much confusion, I'll go ahead and modify the 2 existing classes to directly call the calculator method.

rubensworks:
Sure, modifying the existing recipe type is fine. It would become a bit similar to the crafting recipe type, so we can probably even use the same method name as there.

kirjorjos:
I'm closing this PR to open a new one on a new branch to both keep the commit history cleaner and help me make sure I don't forget to delete any unrequired code.

Issue #1449, opened 2024-12-29
Tooltip Lines Operator


pippinsmith:
#### Issue type:

- :heavy_plus_sign: Feature request 

____

#### Short description:

**Idea:** an operator of the following signature: _Item -> List(String)_; Taking an item and returning a list composed of the lines of the tooltip of the input.

Tooltips are already a "first class citizen" in IntegratedTerminals for search purposes, but can't be used in filters etc.

My reason for needing/wanting this: reading tooltip for the existence of emc values provided by ProjectE, the possibility of parsing the value with regex.

From a quick look at the code, it looks straightforward to implement the functionality of this, so **I'm willing to attempt a PR** if this is something you'd merge, though I'm not yet sure which item I can easily use in the test suite (energy battery maybe?) since my gut feeling doesn't trust the dynamic nature of "<Shift for info"> (which is probably a translation string as well, so maybe harder to test? idk, I'm bad at tests)

rubensworks:
Thanks for the suggestion!

rubensworks:
Good idea, a PR is definitely welcome!

I believe tools such as pickaxes also show some tooltip lines, those could be good testing items.

If you would make a PR, could you target the `master-1.19-lts` branch? I can upmerge it to higher MC versions when releasing.

pippinsmith:
initial attempt: 
![image](https://github.com/user-attachments/assets/05c22aea-70af-4fb9-9fa2-19ee65bcc379)

![image](https://github.com/user-attachments/assets/ff0d456c-d0f9-459b-86c4-23123c28365d)

question: should I remove the item name and/or empty lines or just leave those in for the player to handle (ingame it's only a `slice` or `tail` away)

gonna start writing tests now

rubensworks:
Nice!

> question: should I remove the item name and/or empty lines or just leave those in for the player to handle (ingame it's only a slice or tail away)

I think keeping it as-is is fine. The player will have to do post-processing anyways if they want to do anything useful with it.

Issue #1402, opened 2024-10-10
Support for a  'JSON Text' Type


klugemonkey:
#### Issue type:

- :heavy_plus_sign: Feature request 

____

#### Short description:

Would like to see the implementation of a new 'Raw JSON Text' type similar to the String type that takes a String as an input that will handle the formatting of bold, italic and colors per Minecraft's Raw JSON text format that will render properly on the Display Panel.

This would allow to program Raw JSON Text strings from other strings and variables to make custom display panels that show formatted information.

Right now it appears Display Panels are very limited to Strings that all display in one color and only support newlines.

I feel this feature would be well worth the effort.

rubensworks:
Thanks for the suggestion!

PR #1450, opened 2024-12-30
Added TooltipLines operator

pippinsmith:
Implements #1449

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -1983,7 +1983,7 @@ public final class Operators {
      */
     public static final IOperator OBJECT_ITEMSTACK_ENTITY_TOOLTIP = REGISTRY.register(OperatorBuilders.ENTITY_1_ITEMSTACK_1
             .inputTypes(ValueTypes.OBJECT_ENTITY, ValueTypes.OBJECT_ITEMSTACK)
-            .output(ValueTypes.LIST).symbol("Entity.itemTooltip").operatorName("entityitemtooltip").interactName("entityItemTooltip")
+            .output(ValueTypes.LIST).symbol("entity_item_tooltip").operatorName("entityitemtooltip").interactName("entityItemTooltip")
             .function(variables -> {
                 ValueObjectTypeEntity.ValueEntity a = variables.getValue(0, ValueTypes.OBJECT_ENTITY);
                 ValueObjectTypeItemStack.ValueItemStack itemStack = variables.getValue(1, ValueTypes.OBJECT_ITEMSTACK);
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -11,7 +11,6 @@ import com.google.re2j.PatternSyntaxException;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import lombok.Lombok;
 import net.minecraft.ResourceLocationException;
-import net.minecraft.client.Minecraft;
 import net.minecraft.nbt.*;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.MutableComponent;
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -1968,16 +1968,34 @@ public final class Operators {
             )).build());
 
     /**
-     * Get the tooltip lines of an itemstack.
+     * Get the tooltip of an itemstack in list form.
      */
-    public static final IOperator OBJECT_ITEMSTACK_TOOLTIP_LINES = REGISTRY.register(OperatorBuilders.ITEMSTACK_1_SUFFIX_LONG
-            .output(ValueTypes.LIST).symbol("tooltip_lines").operatorName("tooltiplines").interactName("tooltipLines")
+    public static final IOperator OBJECT_ITEMSTACK_TOOLTIP = REGISTRY.register(OperatorBuilders.ITEMSTACK_1_SUFFIX_LONG
+            .output(ValueTypes.LIST).symbol("tooltip").operatorName("tooltip").interactName("tooltip")
             .function(input -> {
                 ValueObjectTypeItemStack.ValueItemStack itemStack = input.getValue(0, ValueTypes.OBJECT_ITEMSTACK);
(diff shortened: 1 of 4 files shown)


src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1965,6 +1967,19 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
...
 
+    /**
+     * Get the tooltip lines of an itemstack.
+     */
+    public static final IOperator OBJECT_ITEMSTACK_TOOLTIP_LINES = REGISTRY.register(OperatorBuilders.ITEMSTACK_1_SUFFIX_LONG
+            .output(ValueTypes.LIST).symbol("tooltip_lines").operatorName("tooltiplines").interactName("tooltipLines")
+            .function(input -> {
+                ValueObjectTypeItemStack.ValueItemStack itemStack = input.getValue(0, ValueTypes.OBJECT_ITEMSTACK);
+                return ValueTypeList.ValueList.ofList(ValueTypes.STRING,
+                        itemStack.getRawValue().getTooltipLines(Minecraft.getInstance().player, TooltipFlag.Default.NORMAL).stream()
```
rubensworks: `Minecraft` is a client-side-only class, so this will unfortunately crash when used on servers.
But I believe the first param of `getTooltipLines` is `Nullable`, so we can just pass `null` here.

pippinsmith:
I had a feeling it would be that way, wasn't sure if the context would be
changing the output so kept it in (copied from the IntegratedTerminals
query code).
I suppose setting it to null will make the output consistent, but might
create scenarios where testing in a terminal query would produce different
results, though that might not be important.
I'll change it when I get home, running some errands right now

rubensworks:
> but might
create scenarios where testing in a terminal query would produce different
results, though that might not be important.

Indeed, there may be cases where it will be different.
Terminals will run the tooltip query client-side, while ID operators run server-side without a player context.
In theory, ID networks _could_ also run based on a specific player, but this would require some extensive internal changes.

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1232,6 +1232,8 @@
     "operator.integrateddynamics.itemstack.nbt.info": "Get the NBT tag of the given item.",
     "operator.integrateddynamics.itemstack.hasnbt": "Has NBT",
     "operator.integrateddynamics.itemstack.hasnbt.info": "If the item stack has an NBT tag.",
+    "operator.integrateddynamics.itemstack.tooltiplines": "Tooltip Lines",
+    "operator.integrateddynamics.itemstack.tooltiplines.info": "Get the tooltip of the given item in a list of strings.",
```
Jack-McKalling: Would it be better to say "**as** a list of strings"?

pippinsmith:
Good catch, I changed that description 7 times based on the other
operators, became blind for whether or not it was good enough

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1965,6 +1967,19 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
                     itemStack -> !itemStack.isEmpty() && itemStack.hasTag()
             )).build());
 
+    /**
+     * Get the tooltip lines of an itemstack.
+     */
+    public static final IOperator OBJECT_ITEMSTACK_TOOLTIP_LINES = REGISTRY.register(OperatorBuilders.ITEMSTACK_1_SUFFIX_LONG
+            .output(ValueTypes.LIST).symbol("tooltip_lines").operatorName("tooltiplines").interactName("tooltipLines")
```
rubensworks: I wonder if `tooltip` instead of `tooltip_lines` would not be sufficient. (no strong opinion) (same would apply to the other names)

pippinsmith: yea, I was basing it on the ItemStack method name, but makes sense for it to be short right now, since there's no version that returns it as a single object (maybe at some point after #1402 ?)

pippinsmith:
Currently having some discussion about this and my original intended use case on the discord, the option of adding an extra Entity.ItemTooltip(Item) operator has been coined for dealing with the player context, won't be finalizing this PR until all discussion is resolved but will continue testing things.

rubensworks:
Both the entity-based and entity-less variants could be valuable.

pippinsmith:
I didn't add tests yet for the entity variant, I noticed that many of the more complex entity operators are missing tests and might have used that as a basis for not trying yet (it works in game, I promise)

feel free to nitpick about naming, I'm working in jetbrains so it's just a refactor button away

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1965,6 +1967,37 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
...
+                        itemStack.getRawValue().getTooltipLines(null, TooltipFlag.Default.NORMAL).stream()
+                            .map(c -> ValueTypeString.ValueString.of(c.getString()))
+                            .toList());
+            }).build());
+    /**
+     * Get the tooltip of an itemstack in list form, using the provided player entity as the player context.
+     */
+    public static final IOperator OBJECT_ITEMSTACK_ENTITY_TOOLTIP = REGISTRY.register(OperatorBuilders.ENTITY_1_ITEMSTACK_1
+            .inputTypes(ValueTypes.OBJECT_ENTITY, ValueTypes.OBJECT_ITEMSTACK)
+            .output(ValueTypes.LIST).symbol("Entity.itemTooltip").operatorName("entityitemtooltip").interactName("entityItemTooltip")
```
rubensworks:
```suggestion
            .output(ValueTypes.LIST).symbol("entity_item_tooltip").operatorName("entityitemtooltip").interactName("entityItemTooltip")
```

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -11,6 +11,7 @@
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import lombok.Lombok;
 import net.minecraft.ResourceLocationException;
+import net.minecraft.client.Minecraft;
```
rubensworks:
```suggestion
```

rubensworks (approved):

rubensworks:
Thanks @pippinsmith!

pippinsmith:
You're welcome, gonna throw the 1.20 build in the modpack I'm playing when I get the chance.
Not sure how that import stayed there, spotless for sure complained the first time when I moved to null, I guess building for a client run skips spotless 🤔

PR #1451, opened 2024-12-30
Update en_us.json

Jack-McKalling:
Fix misleading Entity Reader "Players" aspect tooltip description


Jack-McKalling:
I confirmed that this aspect indeed only just targets those players that are in the target space, not everyone online.

rubensworks:
Thanks!

PR #1452, opened 2024-12-31
Add recipes for clearing facades

kirjorjos:
Adds the squeezer facade recipes as mentioned in #1424


src/main/java/org/cyclops/integrateddynamics/blockentity/BlockEntitySqueezer.java
```diff
@@ -161,9 +161,10 @@ protected void update(Level level, BlockPos pos, BlockState blockState, BlockEnt
                     Optional<RecipeSqueezer> recipeOptional = blockEntity.getCurrentRecipe();
                     if (recipeOptional.isPresent()) {
                         RecipeSqueezer recipe = recipeOptional.get();
+                        ItemStack oldItem = blockEntity.getInventory().getItem(0);
                         blockEntity.getInventory().setItem(0, ItemStack.EMPTY);
-                        for (RecipeSqueezer.IngredientChance itemStackChance : recipe.getOutputItems()) {
-                            if (itemStackChance.getChance() == 1.0F || itemStackChance.getChance() >= level.random.nextFloat()) {
+                        for (RecipeSqueezer.IngredientChance itemStackChance : recipe.assemble(oldItem)) {
+                        if (itemStackChance.getChance() == 1.0F || itemStackChance.getChance() >= level.random.nextFloat()) {
```
rubensworks: Indentation seems to be wrong here.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/FacadeSqueezeCalculator.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.registries.ForgeRegistries;
+
+public class FacadeSqueezeCalculator {
+
+    public static NonNullList<IngredientChance> getOutputItems(ItemStack inputItem) {
+        ItemStack facadeItemStack = new ItemStack(RegistryEntries.ITEM_FACADE);
+
+        facadeItemStack.setTag(null);
```
rubensworks: This shouldn't be necessary.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/FacadeSqueezeCalculator.java
```diff
@@ -0,0 +1,36 @@
...
+        Either<ItemStack, ItemStackFromIngredient> facadeEither = Either.left(facadeItemStack);
+        IngredientChance facade = new IngredientChance(facadeEither, 1.0F);
+
+        CompoundTag nbt = inputItem.getOrCreateTag();
```
rubensworks: If I understand correctly, the following lines attempt to check if the item is a block?
If so, you could just check if item is instanceof `ItemBlock`. That would simplify this code quite a bit.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSqueezer.java
```diff
@@ -61,6 +61,11 @@ public ItemStack assemble(Container inv) {
         return this.outputItems.get(0).getIngredientFirst().copy();
     }
 
+    public NonNullList<IngredientChance> assemble(ItemStack inputItem) {
+        if (!inputItem.is(RegistryEntries.ITEM_FACADE.asItem())) return getOutputItems();
```
rubensworks: I still think it is possible to create a custom recipe serializer type so that we don't have to hardcode things like this.
Concretely, the serializer could provide a subclass of `RecipeSqueezer` that just overrides this `assemble` method.
The `"type"` entry in the recipe JSON would then refer to this new serializer, e.g. `integrateddynamics:squeezer_facade_unwrap`.

The problem with your current approach is that things become more difficult to maintain like this.

src/main/resources/data/integrateddynamics/recipes/mechanical_squeezer/convenience/mechanical_facade_squeeze.json
```diff
@@ -0,0 +1,19 @@
+{
+  "type": "integrateddynamics:mechanical_squeezer",
+  "item": "integrateddynamics:facade",
+  "result": {
+	"items": [
+	      {
```
rubensworks: Indentation seems a bit off in this file.

src/main/resources/data/integrateddynamics/recipes/squeezer/convenience/facade_squeeze.json
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "integrateddynamics:squeezer",
+  "item": "integrateddynamics:facade",
+  "result": {
+    "items": [
+      {
```
rubensworks: Also here.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSqueezer.java
rubensworks: The helper function in `FacadeSqueezeCalculator` could then be moved to that serializer class.

rubensworks (changes requested):
Thanks @kirjorjos!

src/main/java/org/cyclops/integrateddynamics/blockentity/BlockEntitySqueezer.java
kirjorjos: I definitely missed that, I'll fix that indentation and the JSON file indentations

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/FacadeSqueezeCalculator.java
kirjorjos: I'm not trying to check if it's a block anywhere, just grab the assumed block from the NBT, letting it be an empty ItemStack if malformed or missing NBT as that would be the case in squeezing a blank facade

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSqueezer.java
kirjorjos: I'm fine moving the method from the FacadeSqueezeCalculator to a separate serializer, but I thought the `type` key in the JSON object referenced the recipe type, not the serializer?

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/FacadeSqueezeCalculator.java
kirjorjos: Initially, my testing was giving back facades with empty NBT which didn't stack with the ones already in world and from the other recipes, but testing now, it is in fact unnecessary so I'll remove it.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/FacadeSqueezeCalculator.java
rubensworks: You can probably use this one then: https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.21-lts/src/main/java/org/cyclops/integrateddynamics/item/ItemFacade.java#L41

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSqueezer.java
rubensworks: The type refers to the serializer. For example, the `clear_part.json` recipe refers to the `integrateddynamics:crafting_special_nbt_clear` serializer, which is of the crafting recipe type.

kirjorjos:
I'm not sure if there's a certain way I should let you know that  I believe I've addressed everything you mentioned, but I had pushed 3 commits on February 3rd that for some reason are showing grouped with the ones from December.  Apologies if you had noticed and hadn't gotten to it yet, but wanted to say something as it has been a bit.

rubensworks:
There should be a button to re-request a review from someone. Somewhere on the top-right of this page.

buildAndRunClient.sh
```diff
@@ -0,0 +1,34 @@
+gnome-terminal -- bash -c "
```
rubensworks: This file probably shouldn't be here.

src/main/resources/data/integrateddynamics/recipes/mechanical_squeezer/convenience/mechanical_facade_squeeze.json
```diff
@@ -0,0 +1,25 @@
...
+  "type": "integrateddynamics:facade_squeeze_mechanical",
+  "item": {
+    "item": "integrateddynamics:facade"
+  },
+  "result": {
+    "items": [
+      {
+        "item": {
+          "item": "integrateddynamics:facade",
+		  "chance": 1,
```
rubensworks: Indentation is off here.

src/main/resources/data/integrateddynamics/recipes/squeezer/convenience/facade_squeeze.json
```diff
@@ -0,0 +1,24 @@
...
+  "type": "integrateddynamics:facade_squeeze",
+  "item": {
+    "item": "integrateddynamics:facade"
+  },
+  "result": {
+    "items": [
+      {
+        "item": {
+          "item": "integrateddynamics:facade",
+		  "chance": 1,
```
rubensworks: Here as well

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueeze.java
```diff
@@ -0,0 +1,56 @@
...
+    private IngredientChance facadeItemChance = new IngredientChance(Either.<ItemStack, ItemStackFromIngredient>left(new ItemStack(RegistryEntries.ITEM_FACADE)), 1.0f);
+
+    public RecipeFacadeSqueeze(ResourceLocation id, Ingredient inputIngredient, NonNullList<IngredientChance> outputItems, FluidStack outputFluid) {
+        super(id, inputIngredient, outputItems, outputFluid);
+        this.inputIngredient = inputIngredient;
+    }
+
+    public NonNullList<IngredientChance> assemble(ItemStack inputItem) {
+
+        if (!(inputItem.getItem() instanceof ItemFacade)) {
```
rubensworks: It might be better to just make these cases illegal in the serializer.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueeze.java
```diff
@@ -0,0 +1,56 @@
...
+
+        if (facadeBlockItem == null) return NonNullList.of(facadeItemChance, facadeItemChance);
+        Either<ItemStack, ItemStackFromIngredient> inputItemEither = Either.left(facadeBlockItem);
+        IngredientChance outputItemChance = new IngredientChance(inputItemEither, 1.0f);
+        return NonNullList.of(outputItemChance, outputItemChance, facadeItemChance);
+    }
+
+    @Override
+    public boolean matches(Container inv, Level worldIn) {
+        ItemStack inputCopy = inv.getItem(0).copy();
```
rubensworks: Something like `inputItem.getItem() instanceof ItemFacade` might be more efficient.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueezeMechanical.java
```diff
@@ -0,0 +1,44 @@
...
+import com.mojang.datafixers.util.Either;
+
+import net.minecraft.core.NonNullList;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.item.crafting.RecipeSerializer;
+import net.minecraftforge.fluids.FluidStack;
+
+public class RecipeFacadeSqueezeMechanical extends RecipeMechanicalSqueezer {
```
rubensworks, line 12: Some code duplication here compared to the non-mech recipe.
Maybe we can make some non-trivial methods static and call them from here?

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSerializerFacadeSqueeze.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraftforge.fluids.FluidStack;
+
+public class RecipeSerializerFacadeSqueeze implements RecipeSerializer<RecipeFacadeSqueeze> {
+
+    @Override
+    public RecipeFacadeSqueeze fromJson(ResourceLocation recipeId, JsonObject json) {
+        JsonObject result = GsonHelper.getAsJsonObject(json, "result");
+
+        // Input
+        Ingredient inputIngredient = RecipeSerializerHelpers.getJsonIngredient(json, "item", true);
```
rubensworks: We prob don't need input, as it's always a facade.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSerializerFacadeSqueeze.java
```diff
@@ -0,0 +1,63 @@
...
+
+        // Output
+        NonNullList<RecipeSqueezer.IngredientChance> outputItemStacks = RecipeSerializerSqueezer.getJsonItemStackChances(result, "items");
```
rubensworks: Not sure we even need outputs, as it's also always the facade?
I would guess the serializer to require no special JSON params.

rubensworks (changes requested):
Thanks! This is the direction I had in mind as well :-)

buildAndRunClient.sh
kirjorjos: I was aware I accidentally pushed that, but could've sworn I made another commit minutes after to delete it.  I'll double check that my next commit deletes it.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueeze.java
kirjorjos: I can def move it to the serializer, would you prefer it silently fail to load, or log a warning?

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeSerializerFacadeSqueeze.java
kirjorjos: I had figured this would make it more dynamic in case we added something in the future extending facade, but I can definitely just not check it if you want me to.

rubensworks: Given the name of this serializer, that seems like the right way to go forward.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueezeMechanical.java
kirjorjos: The only non-trivial method I'm seeing here is assemble which I guess could be inherited from RecipeFacadeSqueeze, but I can def make another static if I know which.

rubensworks: 👍

kirjorjos: I'll take that 👍 as make it inherited?

rubensworks: Not sure if inheritance will work here. But if you can, feel free. If not, static is as good.

(Over the years, I came to prefer code reuse through composition instead of inheritance, as the former is more flexible)

kirjorjos: I'll give inheritance a shot as I don't see why it shouldn't work, I can't make it static though as it's overriding a non-static method.

kirjorjos: Once I opened it in eclipse again, I do see why I can't just inherit the method.  I made a static method to call making the assemble method trivial.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueeze.java
```diff
@@ -0,0 +1,50 @@
...
+
+        if (facadeBlockItem == null) return NonNullList.of(facadeItemChance, facadeItemChance);
+        Either<ItemStack, ItemStackFromIngredient> inputItemEither = Either.left(facadeBlockItem);
+        IngredientChance outputItemChance = new IngredientChance(inputItemEither, 1.0f);
+        return NonNullList.of(outputItemChance, outputItemChance, facadeItemChance);
+    }
+
+    @Override
+    public boolean matches(Container inv, Level worldIn) {
+         return (inv.getItem(0).getItem() instanceof ItemFacade);
```
rubensworks, line 44: I suspect this check will fail if the facade is not in the first slot, but for example in the middle slot of the crafting grid.
Could you double-check this?

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueezeMechanical.java
```diff
@@ -0,0 +1,26 @@
...
+import org.cyclops.integrateddynamics.RegistryEntries;
+
+import net.minecraft.core.NonNullList;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.item.crafting.RecipeSerializer;
+import net.minecraftforge.fluids.FluidStack;
+
+public class RecipeFacadeSqueezeMechanical extends RecipeMechanicalSqueezer {
```
rubensworks, line 12: The `matches` override seems to be missing here.

src/main/resources/data/integrateddynamics/recipes/mechanical_squeezer/convenience/mechanical_facade_squeeze.json
rubensworks: Indentation still seems off here.

src/main/resources/data/integrateddynamics/recipes/mechanical_squeezer/convenience/mechanical_facade_squeeze.json
```diff
@@ -0,0 +1,25 @@
...
+          "count": 1
+        }
+      },
+	  {
+        "item": {
+          "item": "integrateddynamics:facade",
+		  "chance": 1,
```
rubensworks: Here as well.

src/main/resources/data/integrateddynamics/recipes/squeezer/convenience/facade_squeeze.json
rubensworks: Also here, and below.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueezeMechanical.java
```diff
@@ -0,0 +1,26 @@
...
+	
+    public RecipeFacadeSqueezeMechanical(ResourceLocation id, Ingredient inputIngredient, NonNullList<IngredientChance> outputItems, FluidStack outputFluid, int duration) {
+        super(id, inputIngredient, outputItems, outputFluid, duration);
+    }
+
+    public NonNullList<IngredientChance> assemble(ItemStack inputItem) {
```
rubensworks, line 19: Does this one actually get called anywhere?
If I'm correct, this is a new method (no override).
But it doesn't seem to get called anywhere.

rubensworks:
Rebasing on master may be required as well, as there seem to be merge conflicts.

Jack-McKalling:
> Rebasing on master may be required as well, as there seem to be merge conflicts.

the thing says this:
> Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

but right above that it says this instead:
> Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not GET 'https://maven.pkg.github.com/CyclopsMC/packages/org/cyclops/commoncapabilities/commoncapabilities/1.19.2-2.9.0-88/commoncapabilities-1.19.2-2.9.0-88.pom'. Received status code 401 from server: Unauthorized

(deepest nested exception)
All of that doesn't sound to me like it's just a GIT operation that would solve it, but I don't know anything.

kirjorjos:
> > Rebasing on master may be required as well, as there seem to be merge conflicts.
> 
> the thing says this:
> 
> > Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
> 
> but right above that it says this instead:
> 
> > Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not GET 'https://maven.pkg.github.com/CyclopsMC/packages/org/cyclops/commoncapabilities/commoncapabilities/1.19.2-2.9.0-88/commoncapabilities-1.19.2-2.9.0-88.pom'. Received status code 401 from server: Unauthorized
> 
> (deepest nested exception) All of that doesn't sound to me like it's just a GIT operation that would solve it, but I don't know anything.

If I'm not mistaking, I believe that's because the github action doesn't have an auth key.

kirjorjos:
> Rebasing on master may be required as well, as there seem to be merge conflicts.

I agree that github is acting weird with it as it isn't even seeing I changed half the files I changed looking now, I haven't really heard of rebasing before, but I'll do some research and give it a try.

src/main/java/org/cyclops/integrateddynamics/core/recipe/type/RecipeFacadeSqueeze.java
kirjorjos: Unless I'm missing something, both squeezer types only have 1 input slot?

rubensworks: Yes, of course, my bad 🤦

kirjorjos:
> > Rebasing on master may be required as well, as there seem to be merge conflicts.
> 
> I agree that github is acting weird with it as it isn't even seeing I changed half the files I changed looking now, I haven't really heard of rebasing before, but I'll do some research and give it a try.

I'm pretty sure I managed to rebase, but I guess if I did something wrong with it, I can try some more.

kirjorjos:
Playing around in game, I am noticing a bug that this PR brings to light: If the facaded item has NBT, that NBT is lost when making the facade.  I think the best way to resolve this would be storing the NBT with the facade, maybe something like `{"block": {"Name": "mod_name:item_name", "tag": {...}}}` as opposed to the current `{"block": {"Name": "mod_name:item_name"}}`?  My only thought against this is that it would then cause any pre-exisiting facades to not be able to stack with new ones.  I could probably add some extra logic to only have the "tag" key if there is data to fix that, but I wanted to run it by you and get your thoughts before doing much work with it.  A good example item to test with is an InDy battery with some energy stored.

src/main/resources/data/integrateddynamics/recipes/special/facade_clear.json
```diff
@@ -0,0 +1,4 @@
+{
```
rubensworks: I guess this file can be removed, since we already have `clear_facade.json`?

rubensworks:
> Playing around in game, I am noticing a bug that this PR brings to light

This issue existed before this PR already, right?
If so, let's just create a separate issue for this.
It shouldn't be included in this PR IMO.

Besides my other comment above, this should be good to go.
So if it's ok from your end, I will test this in-game, and merge.

kirjorjos:
> > Playing around in game, I am noticing a bug that this PR brings to light
> 
> This issue existed before this PR already, right? If so, let's just create a separate issue for this. It shouldn't be included in this PR IMO.
> 
> Besides my other comment above, this should be good to go. So if it's ok from your end, I will test this in-game, and merge.

It did exist before this PR, but it wasn't particularly an issue before this PR as there was no normal instance where the player would see something related to the NBT.   Assuming you still want to make a new issue for this though, I'm good to let you test and merge.

If I'm not mistaking, it should be `Empty Facade -> Empty Facade`, `Filled Facade -> Empty Facade, Facaded Item` on both squeezers and `Empty Facade -> Empty Facade`, `Filled Facade -> Filled Facade` in a shapeless crafting grid.

rubensworks:
The `Empty Facade -> Empty Facade` on both is probably not what we'd want.
Especially for the mechanical squeezer it's a bit counter-intuitive that this is allowed, since it's a no-op, but it will consume energy.
Maybe you could have a look at disallowing that?
After that, I will have a look myself in-game.

kirjorjos:
I can explicitly disable empty -> empty if you want, the empty -> empty is more of a byproduct of allowing any data facade -> empty plus the block if there is one to also handle the case for invalid data. (Eg. If a mod got removed so the block that used to be facaded doesn't exist)

rubensworks:
Removing invalid data makes sense indeed. But if the facade is really just empty, I think we should disallow it.

kirjorjos:
Unless I'm missing something, it looks like I would need to update the [Mechanical]Squeezer block entity class as `matches` is only called on world load with a generic facade.  Do you have any particular place you think would make sense to put it or you want me to just try somewhere where it works?

rubensworks:
> as matches is only called on world load with a generic facade

Are you sure? That's not how recipes are handled based on my recollection.
I thought recipes are always iterated over every time something is crafted (which is why we have a cache in place to optimize these loops).

> Do you have any particular place you think would make sense to put it or you want me to just try somewhere where it works?

The match method should be a good place indeed.
I wouldn't put it in the block entity class in any case.

kirjorjos:
I had thought i remembered matches being called on every run, but it didn't seem to be called when testing with a changed matches, so I launched in debug mode, set a break point, and confirmed it's not called once the world is loaded.

rubensworks:
> but it didn't seem to be called when testing with a changed matches

Probably because of the cache.

kirjorjos:
Testing some more, it seems you are right that cache is making my tests behave weirdly.  To confirm, the cache should reset when I leave the world and rejoin?

kirjorjos:
I'm thinking I must be missing something as from a complete MC restart, after loading into the world, before trying any recipes (so the cache should be empty), I'm debugging matched on FacadeSqueeze, it returns false, but still gets processed.  I also tried using the mechanical squeezer from a fresh MC restart and can't seem to catch inside the matches call in SqueezeFacadeMechanical to see why it's not working.

kirjorjos:
Testing some more, it seems like it's only calling the matches method in RecipeSqueezer and RecipeMechanicalSqueezer, and unless I'm missing something (very possible), it looks to me like the logic for which method to call is inside Cyclopscore?

rubensworks:
> To confirm, the cache should reset when I leave the world and rejoin?

Not sure actually. It will reset in any case on a game restart.

> it seems like it's only calling the matches method in RecipeSqueezer and RecipeMechanicalSqueezer, and unless I'm missing something (very possible)

I don't remember the precise mechanics myself tbh. I'd have to dive into the code myself.

kirjorjos:
I guess for right now, do you want me to push what I'm 99% sure should work with a patched cyclopscore, or did you want to look at it first or...?  Sorry, I'm just unsure what I should be doing here right now.

rubensworks:
I suspect not changes should be needed in CC. But if you think there must be, feel free to open a PR.

Issue #797, opened 2019-12-28
Operator to override/remove NBT from an item/fluid


Pultex:
#### Issue type:Question

Can you remove all NBT data from a entity list or item list but keep all items still in a list ?

____

#### Question:

I wanted to remove the NBT data from all items in an entity reader list variable.
I see that I can convert the whole list into a item list with MAP and entity -> item.
I was hoping to be able to use some thing like MAP NBT -> negation( ) -> item or something. 

I don't think you an pull all items from a active list strip NBT then add them back to a new list
with no NBT, unless I don't see what would allow you to do this.

just a simple answer would do no need to make a list of steps.

rubensworks:
This is not possible yet AFAIK.
Ideally, we should have a `withnbt` operator (similar to `withsize`).

I'll convert this issue into a feature request.

GreyMario:
Try mapping the list to the unique name, then back to an item using operators `String Unique Name` and then `Item Item By Name`. This doesn't provide a method to _override_ NBT, and will remove stack sizes, but will retain most of what identifies an item as an item. Naturally, items which require NBT or capabilities to be uniquely identified will cease to be uniquely identified properly.

TacocaTYT:
perhaps add i as a config option for pack devs including mods that need NBT such as resource hogs from sf4

rubensworks:
This is partially resolved in https://github.com/CyclopsMC/IntegratedDynamics/pull/1463, but I won't close this one yet, as I realized we'll also need a corresponding operator to set the NBT of a fluid.

PR #1463, opened 2025-02-02
added tests and English lang updates

kirjorjos:
Closes issue #797 by adding a with_tag operator

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -1975,7 +1975,7 @@ public final class Operators {
             .function(variables -> {
                 ItemStack inputItem = variables.getValue(0, ValueTypes.OBJECT_ITEMSTACK).getRawValue().copy();
                 ValueTypeNbt.ValueNbt tag = variables.getValue(1, ValueTypes.NBT);
-                inputItem.setTag((CompoundTag)tag.getRawValue().orElse(new CompoundTag()));
+                inputItem.setTag((CompoundTag)tag.getRawValue().orElse(null));
                 return ValueObjectTypeItemStack.ValueItemStack.of(inputItem);
             }).build());
 
--- a/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
@@ -1973,7 +1973,7 @@ public final class Operators {
             .inputTypes(ValueTypes.OBJECT_ITEMSTACK, ValueTypes.NBT)
             .output(ValueTypes.OBJECT_ITEMSTACK).symbol("item_with_tag").operatorName("itemwithtag").interactName("itemWithTag")
             .function(variables -> {
-                ItemStack inputItem = variables.getValue(0, ValueTypes.OBJECT_ITEMSTACK).getRawValue();
+                ItemStack inputItem = variables.getValue(0, ValueTypes.OBJECT_ITEMSTACK).getRawValue().copy();
                 ValueTypeNbt.ValueNbt tag = variables.getValue(1, ValueTypes.NBT);
                 inputItem.setTag((CompoundTag)tag.getRawValue().orElse(new CompoundTag()));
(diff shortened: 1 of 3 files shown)


src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1966,6 +1966,19 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
...
+    /**
+     * Get an itemstack with the given nbt tag
+     */
+    public static final IOperator OBJECT_ITEMSTACK_WITH_TAG = REGISTRY.register(OperatorBuilders.ITEMSTACK_2_LONG
+            .inputTypes(ValueTypes.OBJECT_ITEMSTACK, ValueTypes.NBT)
+            .output(ValueTypes.OBJECT_ITEMSTACK).symbol("item_with_tag").operatorName("itemwithtag").interactName("itemWithTag")
+            .function(variables -> {
+                ItemStack inputItem = variables.getValue(0, ValueTypes.OBJECT_ITEMSTACK).getRawValue();
+                ValueTypeNbt.ValueNbt tag = variables.getValue(1, ValueTypes.NBT);
+                inputItem.setTag((CompoundTag)tag.getRawValue().orElse(new CompoundTag()));
```
rubensworks: Before calling `setTag`, let's take a copy of the stack first, to avoid actual mutations on in-world items.

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1232,6 +1232,8 @@
     "operator.integrateddynamics.itemstack.nbt.info": "Get the NBT tag of the given item.",
     "operator.integrateddynamics.itemstack.hasnbt": "Has NBT",
     "operator.integrateddynamics.itemstack.hasnbt.info": "If the item stack has an NBT tag.",
+	"operator.integrateddynamics.itemstack.itemwithtag": "Item With Tag",
```
rubensworks: Indentation looks a bit off here.

rubensworks (changes requested):
Thanks @kirjorjos!

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
kirjorjos: tried to resolve with a comment, didn't send the comment, I had said that I did test this in game to make sure it didn't, but that I added the .copy() because it won't hurt anything either and if nothing else, makes it clearer that it won't modify the initial item

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1966,6 +1966,19 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
...
+    /**
+     * Get an itemstack with the given nbt tag
+     */
+    public static final IOperator OBJECT_ITEMSTACK_WITH_TAG = REGISTRY.register(OperatorBuilders.ITEMSTACK_2_LONG
+            .inputTypes(ValueTypes.OBJECT_ITEMSTACK, ValueTypes.NBT)
+            .output(ValueTypes.OBJECT_ITEMSTACK).symbol("item_with_tag").operatorName("itemwithtag").interactName("itemWithTag")
+            .function(variables -> {
+                ItemStack inputItem = variables.getValue(0, ValueTypes.OBJECT_ITEMSTACK).getRawValue().copy();
+                ValueTypeNbt.ValueNbt tag = variables.getValue(1, ValueTypes.NBT);
+                inputItem.setTag((CompoundTag)tag.getRawValue().orElse(new CompoundTag()));
```
rubensworks: Now that I look at it again, I would expect that if the optional tag is not present, that no tag is attached to the stack (instead of the empty tag as is done now).
Would you agree?

kirjorjos: I suppose it would make more sense to not have a tag at all.  I shut my PC down tonight, but that'll be an easy fix tomorrow mornning.

rubensworks:
Thanks @kirjorjos!

PR #1471, opened 2025-02-14
Fix what_would_i_be_looking_at advancement

Th3Fanbus:
For some reason, the variable had a spurious `"type": "value_type",` which is not present in the 1.20 version of the mod. So, remove this so that the advancement can be unlocked once again.


rubensworks:
Thanks @Th3Fanbus, but the fix is not entirely correct, as this would make the advancement trigger for all possible variable cards.
I just pushed a slightly different fix that solves the problem.

Th3Fanbus:
No problem, thanks for fixing it!

PR #1474, opened 2025-02-16
Performance optimizations: Pathfinder ~70x faster, also improved general tick speed

Lssikkes:
Before:
image

(tick 12-15 ms, destroying one block 7000 ms stall on server)

After:
image

(tick 4-5 ms, destroying one block 117 ms stall on server)

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementTile.java
+++ b/src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementTile.java
@@ -13,7 +13,7 @@ public class PathElementTile<T extends BlockEntity> extends PathElementCable {
 
     private final T tile;
     private final ICable cable;
-    private DimPos position = null;
+    private final DimPos position;
 
     public PathElementTile(T tile, ICable cable) {
         this.tile = tile;
--- a/src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementTile.java
+++ b/src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementTile.java
@@ -13,10 +13,12 @@ public class PathElementTile<T extends BlockEntity> extends PathElementCable {
 
     private final T tile;
     private final ICable cable;
+    private DimPos position = null;
 
     public PathElementTile(T tile, ICable cable) {
         this.tile = tile;
         this.cable = cable;
+        this.position = DimPos.of(tile.getLevel(), tile.getBlockPos());
     }
 
     protected T getTile() {
@@ -30,7 +32,7 @@ public class PathElementTile<T extends BlockEntity> extends PathElementCable {
 
     @Override
     public DimPos getPosition() {
-        return DimPos.of(tile.getLevel(), tile.getBlockPos());
+        return position;
     }
 
 }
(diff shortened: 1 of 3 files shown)


rubensworks:
Thanks for this @Lssikkes!

It has been on my todo-list for a long time to look into such large-scale performance analysis, but you beat me to it :-)
By the way, did you have an automated way of creating such large networks? If yes, could you share these? As this could be useful for avoiding future performance regressions.

I will have a more detailed look at this PR in the near future.

Lssikkes:
Unfortunately not automated, I just added worldedit and started creating a few large slabs with //set integrateddynamics:cable .. And then pointed VisualVM at it. With worldedit it still lags a lot because for every block placement it does, it will recalculate the whole network. Probably something to be done there that it defers network updates until post tick, and then check if the blocks have already been touched or so, to prevent it from going that wild. But it doesn't happen in normal gameplay too much (except when detonating a TNT/creeper inside the dense cable grid, that'll create similar performance issues), so I don't think it's of the highest priority.

Here's my save:
[TestWorld.zip](https://github.com/user-attachments/files/18822741/TestWorld.zip)

src/main/java/org/cyclops/integrateddynamics/core/network/ConcurrentWorldIngredientsProxy.java
```diff
@@ -107,7 +107,7 @@ public void onWorldTick() {
                 this.states.remove(oldPosition);
                 this.instances.remove(oldPosition);
             }
-            this.oldPositions = Sets.newHashSet(newPositions);
+            this.oldPositions = (Set<PartPos>)newPositions;
```
rubensworks, line 110: Instead of the cast, we can probably change the signature of the `getPositions` method.
Unless you see a specific reason for doing it like this? (not needed for backwards-compat, as no other mods are calling `getPositions`)

rubensworks (changes requested):
I just have one possible suggestion. Everything else looks perfect to be merged!

src/main/java/org/cyclops/integrateddynamics/core/network/ConcurrentWorldIngredientsProxy.java
Lssikkes: Yeah that would also work, might be better indeed

src/main/java/org/cyclops/integrateddynamics/capability/path/PathElementTile.java
```diff
@@ -13,10 +13,12 @@ public class PathElementTile<T extends BlockEntity> extends PathElementCable {
 
     private final T tile;
     private final ICable cable;
+    private DimPos position = null;
```
rubensworks:
```suggestion
    private final DimPos position;
```

src/main/java/org/cyclops/integrateddynamics/core/network/ConcurrentWorldIngredientsProxy.java
rubensworks: Is it possible you forgot to push your change?

rubensworks:
@Lssikkes Did you see my [last comment](https://github.com/CyclopsMC/IntegratedDynamics/pull/1474#discussion_r1966719412) above?

rubensworks:
Nvm, I just applied the change myself.

Thanks again for the PR, I'll include it in the next release!

PR #1476, opened 2025-02-21
Small fixes to the en_us lang file

Jack-McKalling:
Some small fixes to the base language file:

- Remove reference to "crystallised menril resin"
Producing crystallised menril means turning resin into a block, so the produce can't be called resin anymore.
So lets just call it generalized "crystallised menril", since you get a block of it but you can convert it back and forth into chunks for free so there isn't a exclusive form to it.

- Fix tooltip typo in static light panel

- Fix mention of "degrees Kelvin"
The Kelvin unit is not actually called "degrees Kelvin", it's just "Kelvin".

- Fix "raining" grammar sentence in aspect
There is an obsolete word in "If there it is raining in the world".


rubensworks:
Thanks!

PR #1516, opened 2025-05-31
Fix typos in language file

Jack-McKalling:
Just noticed an older minor grammar issue. Added 1 letter to fix it.
I'm going to use this draft to suggest more fixes later, to make it more valuable.


Jack-McKalling:
When I'm done, I suggest to squash all commits into one. They all merely touch one single file anyway.

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1042,7 +1042,7 @@
     "operator.integrateddynamics.relational.gt": "Greater Than",
     "operator.integrateddynamics.relational.lt": "Less Than",
     "operator.integrateddynamics.relational.notequals": "Not Equals",
-    "operator.integrateddynamics.relational.notequals.info": "Type one and two must be equal.",
+    "operator.integrateddynamics.relational.notequals.info": "Type one and two must not be equal.",
```
Jack-McKalling: I suspect I made an error here. I guess this statement applies to the _type_ of the inputs, which must be the same for both equals and notequals operators.

Please confirm.

rubensworks: Not just the type, but the actual *value* must be not equal.

rubensworks: I'm surprised by this.
Can you check in-game if the types must really be the same?

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1506,7 +1506,7 @@
     "operator.integrateddynamics.nbt.from_long.info": "Create an NBT Long tag from the given Long value",
     "operator.integrateddynamics.nbt.from_double": "NBT Double From Double",
     "operator.integrateddynamics.nbt.from_double.info": "Create an NBT Double tag from the given Double value",
-    "operator.integrateddynamics.nbt.from_float": "NBT Float From Double",
+    "operator.integrateddynamics.nbt.from_float": "NBT Double From Float",
```
Jack-McKalling: Not 100% certain about this consistency fix. Did I understand this correctly?

src/main/resources/assets/integrateddynamics/lang/en_us.json
Jack-McKalling: Yep, they indeed need to be the same, as I suspected. See my test (both cards have value `1`):
image

Jack-McKalling: I'll revert this change in the next commit, but I guess both descriptions might need further text to explain what they actually do.

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1815,22 +1815,22 @@
...
     "info_book.integrateddynamics.manual.parts.settings.text3": "The Ticks/Operation allows you to configure the ticking frequency of this part. The higher, the slower it operates.",
     "info_book.integrateddynamics.manual.parts.settings.text4": "The Priority determines the order in which this part is executed within a single network tick.",
     "info_book.integrateddynamics.manual.parts.settings.text5": "The Energy Channel indicates the channel from which energy should be consumed when this part ticks. This is only applicable if energy consumption for networks is enabled.",
 
     "info_book.integrateddynamics.manual.parts.offsets": "Offsets",
-    "info_book.integrateddynamics.manual.parts.offsets.text1": "By default, parts will target the direct neighbouring position. However, when &lPart Enhancements&r are applied, the target position can be changed through the part's &lOffset&r screen.",
+    "info_book.integrateddynamics.manual.parts.offsets.text1": "By default, parts will target the direct neighboring position. However, when &lPart Enhancements&r are applied, the target position can be changed through the part's &lOffset&r screen.",
```
Jack-McKalling, line 1831: It pains me to have to do this, but this is `en_us` after all...

src/main/resources/assets/integrateddynamics/lang/en_us.json
Jack-McKalling: With help from expert Miner, I now understand that I fixed this incorrectly. This name was correct, but the description is wrong and so is the string ID itself. But that's because "from_double" already exists. This string should really be renamed to "float_from_double" instead, but that's going to have to be separate from this PR.

I'll revert this change and correct the description instead in next commit.

Jack-McKalling:
@rubensworks I've finished all above changes, could you resolve them upon confirmation?

If you need the commits squashed, and if it would be easier for you, I could re-do this PR in a single commit now that I've completed the list of changes. Let me know.

rubensworks:
Thanks, that's not necessary, I can easily squash here in the GitHub UI.

rubensworks:
Thanks @Jack-McKalling!

PR #1538, opened 2025-07-22
Adds a config to disable the collision for cable

Tonywww2:
In some cases (when there are really a lot of cables in the view), calculating the collision for the cable can be slow. 
So I added a config that allows the player to disable the collision calculation. 

Here is a comparison with about 20000~ cables in a super flat world:
2ca8ef265633ee3cda39e17fbe538665
d1d4ca99fdcea308b3912fd65fbba1bd


rubensworks:
Thanks for the contribution!

PR #1548, opened 2025-09-15
Convert the Fluid Reader's Total Capacity & Amount aspects to `Long` (fixes #1547)

### LINKED ISSUE #1547: Fluid Reader aspects "Total Capacity" and "Total Amount" overflow their limit.
#### Issue type:

- :bug: Bug 

____

#### Short description:

Each block has the ability to expose multiple fluid tanks (up to the integer limit). Every tank can have capacity and amount up Integer.MAX_VALUE. The maximum total fluid capacity/amount a single block can have is Integer.MAX_VALUE * Integer.MAX_VALUE, which is to say that when a block has a total capacity exceeding Integer.MAX_VALUE, the reader's total capacity/amount values will have an integer overflow issue.

1. Place any block that has a total capacity over the integer limit.
2. Observe as the reader shows an incorrect value such as -1 or similar for "Total Capacity" or "Total Amount" aspects.

#### Expected behaviour:

The reader should show the correct total capacity/amount.

____

#### Versions:

- This mod: 1.27.9
- Minecraft: 1.21.1
- Mod loader version: NeoForge 21.1.208

#### Suggested Fix:

I suggest changing the Total Capacity and Total Amount aspects to use Long instead, which will be more than enough to handle the maximum possible value of Integer.MAX_VALUE * Integer.MAX_VALUE total capacity/amount
Relevant code:
[Total Capacity](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.21-lts/src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java#L336-L343)
[Total Amount](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.21-lts/src/main/java/org/cyclops/integrateddynamics/part/aspect/Aspects.java#L324-L331)

#### Workarounds
There is a list of "Tank Fluids" and a list of "Tank Capacity", so you can map them over a cast to long and then reduce the list over add

ton185:
This changes the type of the Fluid Reader's total capacity and total amount aspects to Long instead of Integer, closing #1547

Builds fine, I also went in a world and tested it and it worked fine.

Edit: I did some further testing, making a world without this change and then "upgrading" to the version in this PR. This will break any existing cards that use the total capacity / total amount cards.


Jack-McKalling:
Does this really fix the issue though? It only replaces the INT type to LONG, and doesn't introduce an exact calculation at the limit. So I suppose the same issue still exists, but now at Long.MAX_VALUE.

rubensworks:
Thanks for the PR @ton185!
But as has been remarked, changing value types of aspect like this does indeed imply a breaking change.

I'm open to this change for the next major MC update, but for now, these value types need to remain as-is.

I did recently notice and fixed a similar in Integrated Mekanism, so I have another workaround available, but for now it will remain in the INT range.

PR #1591, opened 2026-01-10
Fix cable shape cache having cache collisions

Archengius:
When there are two cables with different shapes but identical connections and the same parts (but placed on different sides), cache collision will occur, with both shapes having the same cache key, but different collision shapes. This results in ghost collisions when playing in single player and collision desync between server and the client on dedicated servers.

Example of two cable blocks having different shapes but same cache key, resulting in them having the same shape. Notice how on the second part the collision box on the left should not logically exists, but is shown due to the bug mentioned.

image
image


rubensworks:
Oh, good catch, I never even noticed this was a problem 😅

PR #1594, opened 2026-01-18
Fix part capabilities not being invalidated when network is reinitialized

Archengius:
This causes capability queries on part blocks to return old capabilities that are no longer valid to access the network

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/capability/partcontainer/PartContainerTileMultipartTicking.java
+++ b/src/main/java/org/cyclops/integrateddynamics/capability/partcontainer/PartContainerTileMultipartTicking.java
@@ -32,6 +32,7 @@ public class PartContainerTileMultipartTicking extends PartContainerDefault {
     @Override
     protected void setChanged() {
         getTile().setChanged();
+        getLevel().invalidateCapabilities(getPos());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


rubensworks:
Hi @Archengius, thanks for the PR.

Before we go any further we this, could you explain what the exact problem is you are trying to solve?
AFAICS, there are no open/known issues related to capability invalidation not working correctly.
It's important for me to have a reproducible example I can test the fix against.

Also, I thought we already had some cap invalidation logic in place (can't check right now, so I might misremember), so it might be fixable in that place if there is indeed a problem.

Archengius:
Passive interaction does not work with mods that cache capabilities because ID capabilities change without being invalidated, causing mods that look them up to use stale capability objects and be unable to insert into/extract from ID network. Simplest repro I have is an EnderIO alloy smelter with crafting tunnel attached to it for a craft, and the smelter is configured to auto output the result into the crafting tunnel. Then if you try to order the craft it works, but if you invalidate the network in any way (say, by placing or removing a cable next to the interface), the auto output stops working until either block is replaced.

There is 0 logic for capability invalidation that I have found.

rubensworks:
I see, thanks for the additional info.

I'll have to go through this myself, to make sure there are no unintended consequences due to this, and if cap invalidation in all of these places is really necessary.
The change may be small code-wise, but it's quite a significant change.
So we have to make sure there are no unintended consequences.

And indeed, it looks like we don't have any cap invalidation logic in here. I must have been confusing it with another mod.

rubensworks:
@Archengius I just pushed a slightly different change that resolves the described problem on my end.
The change is a bit more locally, with a lower chance of unintended consequences.
Could you try out the latest dev build to confirm it also fixes all issues on your end?
The dev build should land on curseforge as beta soon.

rubensworks:
A new release is out with the above mentioned fix.

PR #1615, opened 2026-02-24
Document raw item/fluid equality ignore quantities

### LINKED ISSUE #1614: Documentation of equality operators unclear regarding data types that contain quantities
#### Issue type:

- :question: Question

____

#### Question:

Per the documentation for the `==` operator:
> Tests whether two values are equal. The types of both must be the same.

Per the documentation for the Raw Item Equals operator:
> If the raw items of the given items are equal, ignoring NBT and damage values.

Per the documentation for the Raw Item Equals No NBT operator:
> If the raw items of the given stacks are equal, ignoring data components but including damage values.

So, what am I supposed to use if I want to check whether two `Item` objects represent the same item (considering NBT and damage values), but possibly different quantities of the item? As written, `==` is IMO not obviously the right answer (it may be the right answer, but it’s not clear/obvious to me) because I would interpret `==` as returning whether two values of type `Item` are equal, and a value of type `Item` includes a quantity, so two values of type `Item` with different quantities are not equal. But the other two operators are too loose if I *do* care about damage and/or NBT values.

The same question applies to fluids, since values of type `Fluid` also include quantities.

Thanks!

Hawk777:
Fixes GH-1614


rubensworks:
Thanks!

PR #1691, opened 2026-07-30
Fix client crash when rendering a facade on a cable with a null model…

mashen456:
… baker

CableModelBase.MODEL_BAKER is populated lazily when the cable item model is baked (ItemModelCable.Unbaked#bake). Under deferred/lazy model baking, e.g. ModernFix's dynamic_resources, it may still be null when a chunk section containing a facade is first meshed, causing a NullPointerException in FaceBakery#bakeQuad (ModelBaker.interner()).

Guard against this in addFacadeQuad by skipping the facade quad while the baker is unavailable; it renders once the item model has been baked.

### DIFF
--- a/src/main/java/org/cyclops/integrateddynamics/client/model/CableModelBase.java
+++ b/src/main/java/org/cyclops/integrateddynamics/client/model/CableModelBase.java
@@ -185,6 +185,13 @@ public abstract class CableModelBase extends DelegatingDynamicItemAndBlockModel 
     }
 
     private void addFacadeQuad(List<BakedQuad> quads, BakedQuad originalQuad, float u0, float v0, float u1, float v1, Direction side) {
+        // MODEL_BAKER is populated lazily when the cable *item* model is baked (see ItemModelCable.Unbaked#bake).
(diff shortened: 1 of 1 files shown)


src/main/java/org/cyclops/integrateddynamics/client/model/CableModelBase.java
```diff
@@ -185,6 +185,13 @@ public List<BakedQuad> getFacadeQuads(BlockStateModel facadeModel, BlockState bl
     }
 
     private void addFacadeQuad(List<BakedQuad> quads, BakedQuad originalQuad, float u0, float v0, float u1, float v1, Direction side) {
+        // MODEL_BAKER is populated lazily when the cable *item* model is baked (see ItemModelCable.Unbaked#bake).
+        // Under deferred/lazy model baking (e.g. ModernFix's dynamic_resources) that may not have happened yet when a
+        // chunk section containing a facade is first meshed, leaving MODEL_BAKER null and crashing FaceBakery#bakeQuad.
+        // Skip the facade quad in that case instead of crashing; it will render once the item model has been baked.
+        if (MODEL_BAKER == null) {
```
rubensworks, line 192: This is not a solution, since this will just make facades invisible.

Issue #1716, opened 2026-09-03
Create Simulated crash/stall when breaking pipes


RopTamen:
#### Issue type:

- :bug: Bug 

____

#### Short description:

When breaking ID pipes on assembled simulated contraptions, this occasionally results in a server stall that leads to a crash. This is on the newest 1.20.1 versions. I'm unsure which mod is technically at fault, but given no other pipes cause a similar issue I'm assuming it's more on this side.

#### Steps to reproduce the problem:

Assemble any contraption with pipes or place them on afterwards and attempt to break any of them. Occasionally the pipe connection will fail to update and the server will stall. This doesn't always happen, so I don't know what specific arrangement causes it, and it may be more or less random or inconsistent.

1. ...

#### Expected behaviour:

The server to not stall and crash when breaking pipes.

____

#### Versions:

integrateddynamics-1.21.1-neoforge-1.35.0-1995
integratedterminals-1.21.1-neoforge-1.8.0-841
integratedtunnels-1.21.1-neoforge-1.10.0-736
integratednbt-1.21.1-neoforge-1.6.2-54
integratedmekanism-1.21.1-neoforge-1.1.0
integratedscripting-1.21.1-neoforge-1.0.26
integratedcrafting-1.21.1-neoforge-1.5.0-646

- This mod: 1.20.1
- Minecraft: 1.20.1
- Mod loader version: NeoForge 21.1.248 

#### Log file:

Too large for one paste, I don't know which parts are important for this issue so this is all of it:
https://pastes.io/cbpngF6Q
https://pastes.io/DDRvdWN1
https://pastes.io/PILpSG3h
https://pastes.io/rfpAaG3e

rubensworks:
Thanks for reporting!

PR #1717, opened 2026-09-03
Fix cables staying behind in their network when moved by contraption mods

rubensworks:
Refs #1716

## The bug

Contraption mods built on Create remove the block entity *before* setting the block to air when they pick up a block. `Contraption#removeBlocksFromWorld` does:

```java
world.removeBlockEntity(add);
int flags = Block.UPDATE_MOVE_BY_PISTON | Block.UPDATE_SUPPRESS_DROPS | Block.UPDATE_KNOWN_SHAPE
    | Block.UPDATE_CLIENTS | Block.UPDATE_IMMEDIATE;
world.setBlock(add, Blocks.AIR.defaultBlockState(), flags);
```

By the time `BlockCable#onRemove` runs, the network can not be reached through the position's `INetworkCarrier` capability anymore (the carrier is either gone, or is the removed block entity's carrier with a `null` network). `CableHelpers#onCableRemoving` hits neither branch and silently does nothing, so the position stays behind in its network forever.

`onCableRemoved` then builds a correct new network for the surviving cables, so **every contraption assembly leaks one network** that is persisted and ticked. On the next world load this surfaces as exactly the warnings from #1716:

```
Skipped loading part from a network at position BlockPos{...} because it has no valid path element.
Detected network position at position BlockPos{...} with corrupted network, recreating network...
```

## Reproduction

Verified in a dev client with Create 6.0.10 and IntegratedDynamics-Compat, driven with clientdevbridge-cli: a mechanical bearing under a row of three cables, toggled with a creative motor.

Before:

```
before:    net#242117713  cables=3 [0/6/3(cable), 1/6/3(cable), 2/6/3(cable)]
assembled: net#1026384218 cables=2 [1/6/3(cable), 2/6/3(cable)]
        || net#242117713  cables=1 [0/6/3(AIR)]        <- orphan
disasm:    net#1649932268 cables=3 [0/6/3, 1/6/3, 2/6/3]
        || net#242117713  cables=1 [0/6/3]             <- same position in two networks
+5 toggles: 4 orphaned networks, all claiming the same dead position
```

After:

```
before:    net#1696949471 cables=3 [0/6/3, 1/6/3, 2/6/3]
assembled: net#1909876647 cables=2 [1/6/3, 2/6/3]
disasm:    net#1099762612 cables=3 [0/6/3, 1/6/3, 2/6/3]
+6 toggles: still exactly one network
```

Reloading the world after the "before" run logs the `Skipped loading part from a network` warnings; after the fix it logs none.

## The change

- `CableHelpers#onCableRemoving` falls back to a position-based lookup when the network can not be reached through the carrier.
- `NetworkHelpers#removeStalePathElement` finds every stored network that still holds the position and removes it. Path elements are compared by position only, so the new `PathElementPosition` is enough to look one up without a block entity.
- `Network#removePathElement` falls back to the network elements it derived for the position earlier when the `INetworkElementProvider` capability is gone as well, so those are cleaned up instead of leaking. Previously it removed the path element from the cluster but skipped element removal and `onNetworkChanged()`, and returned `false`.

## Testing

`testNetworkTwoRemovedWithoutBlockEntity` in `GameTestsNetwork` removes a cable the way contraption mods do and asserts the position is not left behind in any network. It fails on `master-1.21-lts` (`Removed cable is still present in a network`) and passes with the fix.

`./gradlew build` and `./gradlew runGameTestServer` (all 942 tests) pass.

## Not covered by this change

The server stall in the log files attached to #1716 is a separate problem and is not fixed here. All three watchdog thread dumps show the same stack, with no Integrated Dynamics frames in it:

```
ServerChunkCache.getChunk <- Level.getChunk <- lithium ChunkAwareBlockCollisionSweeper
<- Shapes.collide <- Entity.collide <- sable$collideRedirect <- Entity.move <- ItemEntity.tick
```

Right before that tick, Sable logs `Enormous local sub-level collision bounds, quitting.` ~45 times. That guard trips on the entity's own swept AABB in world space, so the item entity already had a huge delta movement before Sable transformed anything; the vanilla collision sweep then walks chunk sections at sub-level coordinates around x=20,481,000 and generates ungenerated chunks synchronously. That belongs to the physics/collision side rather than to this mod.

https://claude.ai/code/session_01PWtMBWaxBPR99VhV5nhkF7

---
_Generated by [Claude Code](https://claude.ai/code/session_01PWtMBWaxBPR99VhV5nhkF7)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ logs/
 # Ignore files specific to dev environments
 secrets.properties
 changelog.txt
+
+# ClientDevBridge session state (golden images are intentionally kept)
+.clientdevbridge/*
+!.clientdevbridge/golden/
--- a/src/integrationtest/java/org/cyclops/integrateddynamics/gametest/GameTestsNetwork.java
+++ b/src/integrationtest/java/org/cyclops/integrateddynamics/gametest/GameTestsNetwork.java
@@ -10,18 +10,25 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 import net.minecraft.world.level.GameType;
+import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.phys.BlockHitResult;
 import net.neoforged.neoforge.gametest.GameTestHolder;
 import net.neoforged.neoforge.gametest.PrefixGameTestTemplate;
 import org.cyclops.cyclopscore.client.particle.ParticleBlurData;
+import org.cyclops.cyclopscore.datastructure.DimPos;
+import org.cyclops.integrateddynamics.IntegratedDynamics;
 import org.cyclops.integrateddynamics.Reference;
 import org.cyclops.integrateddynamics.RegistryEntries;
 import org.cyclops.integrateddynamics.api.network.INetwork;
+import org.cyclops.integrateddynamics.api.path.ISidedPathElement;
+import org.cyclops.integrateddynamics.capability.path.PathElementPosition;
+import org.cyclops.integrateddynamics.capability.path.SidedPathElement;
 import org.cyclops.integrateddynamics.core.helper.CableHelpers;
 import org.cyclops.integrateddynamics.core.helper.NetworkHelpers;
 import org.cyclops.integrateddynamics.core.helper.PartHelpers;
 import org.cyclops.integrateddynamics.core.part.PartTypes;
+import org.cyclops.integrateddynamics.core.persist.world.NetworkWorldStorage;
 
 import java.util.Optional;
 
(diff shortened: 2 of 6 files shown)


rubensworks:
Followed up on whether the first commit regressed performance. It did, and it is fixed in 24405d9.

Measured with a temporary game test that removes a 10×10×10 cable structure the way `Contraption#removeBlocksFromWorld` does, on a game test world that had ~4400 networks stored by that point:

| | total | of which in `removeStalePathElement` | full scans |
|---|---|---|---|
| `master-1.21-lts` | 3689 ms | n/a | n/a |
| 226b3d3 (scan all networks) | 4296 ms | **1741 ms** | 1000 / 1000 |
| 24405d9 (neighbours first) | 2198 ms | **7 ms** | 12 / 1000 |

Walking every stored network is O(networks) per moved block, and a contraption mod triggers it for every block it moves, so it scaled exactly the wrong way for the case this PR is about. The network being looked for is the one the neighbouring cables are still in, so those six positions are checked first; the full scan stays as the fallback for when no neighbour points at it (the last block of a structure, an isolated cable), which was 12 of 1000 removals here.

Two other things that came out of measuring rather than reasoning, in case they are useful when reviewing:

- Instrumenting the whole game test suite, the fallback runs 12 times across all 942 tests on top of the new regression test, all in the TNT/explosion tests. Those are the same class of situation and the removals there are harmless (nothing was found to remove in 11 of them).
- I had first gated the fallback on `world.getBlockEntity(pos) == null`, on the assumption that the block entity is gone at that point. It is not: `getBlockEntity` still returns one during `onRemove` even in the contraption case, so that guard silently disabled the fix while the regression test still passed for the wrong reason. It is not in the pushed code.

`./gradlew build` and `./gradlew runGameTestServer` (942 tests) pass, and the Create mechanical bearing assemble/disassemble cycle is still clean over 8 toggles with the neighbour lookup in place.

---

Issue #239, opened 2017-08-13
[1.12] Unstable Variable Store


Grygon:
I used ID to automate Actually Additions' canola oil upgrade process, but I'm having trouble with it semi-randomly deactivating. Here's a picture of my setup:

![2017-08-12_20 06 42](https://user-images.githubusercontent.com/647846/29245049-28b3d60e-7f9a-11e7-8798-5e0600256018.png)

In short, it reads the fluid block and if its name is "Refined Canola Oil" it sends a redstone signal to the precision dropper, who drops a canola seed. When the block's name is "Crystallized Oil", it sends a redstone signal to the fluid collector who picks it up.

Normally it runs fine, but occasionally when going AFK for a long period of time I come back to no power in my base. The setup is sitting there with the refined canola oil, untriggered to start. In order to start the process again, I have to break the variable store and re-place it, along with inserting all the variable card again--I haven't been able to find another method that works.

I have a similar issue when quitting and re-entering my world, although in this case the readers/writers show an error that the variable could not be found. The same solution (breaking and replacing the variable store) fixes this.

I'm running 0.8.4 on the All the Mods 3 modpack, with Optifine added.

rubensworks:
Sound like a chunk-loading-order-related issue, I'll look into it.

Grygon:
In case it helps, the entire mechanism is located within the same chunk.

rubensworks:
Hmm, then something else must be causing it...

maham:
I'm having something related. Whenever I enter the game I have to break replace my variable storages. Otherwise my setup just spits all my seeds out and ignores my redstone writers. An other way I can fix it is by taking out the variable card and reinserting it in each of the redstone writer in the system.
I have the problem in my automated smeltery too...
Now that I think about it I think all redstone readers/writers are having trouble. The smeltery have a lever to enable molten metal to flow from the smeltery to a tinker tank. I also have a lever to enable flow from the tinker tank to an casting table with an ingot cast. And when I start the game I have to remove->insert the variable cards (or break->place the variable storage) for the levers to work.

MC 1.12.1
Integrated Dynamics 0.8.4
Integrated Tunnels 1.2.5
Forge 14.21.1.2443

rubensworks:
Might be related to #259 

Also, might be related to an "empty tile entity" problem I had before, during world (un)loading.

rubensworks:
I haven't been able to reproduce this issue, but the latest change _should_ resolve it.

[...]

Issue #238, opened 2017-08-08
[1.10, unsure of other versions] Chisels & Bits bounding box on writers


LemmaEOF:
![2017-08-07_23 38 57](https://user-images.githubusercontent.com/6722691/29059469-878ebe56-7bca-11e7-92c3-fe7b4bf3f45c.png)
I'm trying to hide some effects writers in my base with Chisels and Bits, but apparently the empty space on the face with the writer on it isn't somewhere bits can fit. It's kind of dissonant with the writer's model itself. Is fixed in later versions and/or is it intentional? I'm on 1.10.2-0.7.8 right now, for reference.

rubensworks:
This is indeed intentional at the moment, as more complicated bounding boxes would reduce performance. Perhaps I can come up with a clean solution in the future.

LemmaEOF:
Ok, thank you. Since I'm using effect writers, I can just use X- and Y- offset to put them off to the side for now.

Jack-McKalling:
This is also observable when using cable facades. And not just with writers but also with omni-directional connectors.
Since there is a portion around the connector that's transparent within its bounding box, that portion allows you to see through the cable facade block.

PR #1725, opened 2026-09-06
Hash a prototype once per index position update

rubensworks:
`IngredientPositionsIndex.addPosition` and `removePosition` each look a prototype up and then store or drop it. Every ingredient map call builds its own key wrapper, so the prototype was hashed twice for one update. For ItemStack that hash walks a data component map, and `PositionedAddonsNetworkIngredients.applyChangesToChannel` runs each change twice over, once for its own channel and once for the wildcard channel, so it multiplies.

Both now use the `compute` method added in CyclopsMC/CyclopsCore#239, which hashes once.

Behaviour is unchanged. `removePosition` still drops the entry when its position set runs empty and still drops the priority level when the map runs empty; the only difference is that the level check now runs unconditionally, which can only be true when an entry was actually dropped.

## Dependency is now satisfied

CyclopsMC/CyclopsCore#239 is merged and released. `cyclopscore_version` is bumped to `1.30.4-1135` in this PR, which is the CyclopsCore build carrying both #238 and #239 (run 1135, all deploy steps green).

Verified rather than assumed:

- The artifact is live on GitHub Packages. `cyclopscore-26.1.2-neoforge:1.30.4-1135` returns 200; a made-up neighbouring version returns 404, so the check distinguishes.
- The local Maven copy was deleted and the build re-run with `--refresh-dependencies`, so resolution had to go to GitHub Packages. It did, and `./gradlew build` passed.
- `./gradlew runGameTestServer` against that resolved dependency: all 964 required tests pass.

The branch is rebased onto current `master-26-lts`, so it carries the benchmarks from #1722.

## Numbers

`PERFORMANCE_BENCHMARK_ENABLED=true ./gradlew runGameTestServer`, ms per operation, medians of six runs each, alternating between the two configurations within one session so session drift falls on both equally. Both sides carry CyclopsMC/CyclopsCore#238, since this only pays off once hashing costs something.

These were measured before the release, against a locally built CyclopsCore carrying #239. They were **not** re-measured against `1.30.4-1135`; that build is the same code, and the run after the bump was a correctness check, not a benchmark.

| benchmark | without compute | with compute | factor |
|---|---:|---:|---|
| index_modification_single_item | 0.001162 | 0.000688 | **41% faster** |
| index_modification_few_items | 0.001084 | 0.000726 | **33% faster** |
| index_modification_heavy_components | 0.002510 | 0.002077 | **17% faster** |
| index_modification_mixed | 0.000713 | 0.000667 | 6% faster |
| index_modification_plain | 0.000530 | 0.000507 | 4% faster |
| index_modification | 0.001246 | 0.001381 | 11% slower |

Lookups land within 10% either way with no consistent direction, as expected: they do not read-then-write.

The win is concentrated where hashing dominates a modification, meaning many component variants of one item or a large component payload. On the realistic `mixed` shape it is 6% with overlapping distributions, because CyclopsMC/CyclopsCore#238 already skips the component hash for stacks carrying no patch and most instances there carry none. The spread-shape `index_modification` row runs first in the sequence and absorbs warm-up; its two distributions overlap almost completely, so I would not read a regression into it and cannot rule one out either.

No changelog entry, as requested.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mxjin31W1Lmq5XK1CCe84v)_

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -31,6 +31,6 @@ org.gradle.configuration-cache=true
 org.gradle.configureondemand=true
 
 # Dependencies
-cyclopscore_version=1.29.5-1012
+cyclopscore_version=1.30.4-1135
 integrateddynamicscompat_version=26.1.2-neoforge:1.0.0-185
 commoncapabilities_version=2.11.4-344
--- a/src/main/java/org/cyclops/integrateddynamics/core/network/IngredientPositionsIndex.java
+++ b/src/main/java/org/cyclops/integrateddynamics/core/network/IngredientPositionsIndex.java
@@ -138,11 +138,10 @@ public class IngredientPositionsIndex<T, M> implements IIngredientPositionsIndex
         }
 
         T prototype = getPrototype(instance);
-        ObjectOpenHashSet<PartPos> set = positionsMap.get(prototype);
-        if (set == null) {
-            set = new ObjectOpenHashSet<>();
-            positionsMap.put(prototype, set);
-        }
+        // Computed in one call so the prototype is hashed once rather than once to look it up and
+        // once to store it. Hashing a prototype dominates this method for component-bearing types.
+        ObjectOpenHashSet<PartPos> set = positionsMap.compute(prototype,
+                (key, existing) -> existing == null ? new ObjectOpenHashSet<>() : existing);
 
         if (set.add(pos.getPartPos())) {
             Object2IntOpenHashMap<PartPos> positions = this.prioritizedPositions.get(priority);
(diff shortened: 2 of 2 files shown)


rubensworks:
A `master-1.21-lts` version of this now exists as #1728, since the plan is to upmerge rather than maintain both.

`IngredientPositionsIndex.java` is byte identical between the branches, so it is the same patch. I did not re-run the benchmarks on 1.21 and said so there rather than reusing these numbers silently.

Same dependency situation applies there: it needs CyclopsMC/CyclopsCore#241 merged, released and pinned before it compiles.

This PR stays open. Merge whichever branch suits your upmerge direction.

---

PR #1738, opened 2026-09-20
Update to Minecraft 26.3

rubensworks:
Updates IntegratedDynamics from 26.2 to 26.3. `./gradlew build` passes and all **1001 game tests** pass locally, and a dev client was driven on 26.3 with [clientdevbridge](https://github.com/CyclopsMC/clientdevbridge-cli) to exercise the behavioural changes.

## IntegratedDynamics-Compat is temporarily unshadowed

`cyclopscore_version` (`1.30.0-1159`) and `commoncapabilities_version` (`2.11.8-398`) are the published 26.3 builds and resolve.

IntegratedDynamics-Compat is the remaining problem, and it is circular: it compiles against this mod's `slim` artifact, while this mod shadows the compat jar, so neither can build for a new Minecraft version until the other has published. To break it, the compat `shadow` entry in `build.gradle` is commented out with a TODO, so this mod builds and publishes for 26.3 first. CyclopsMC/IntegratedDynamics-Compat#44 can then build against it and publish, after which a follow-up bumps `integrateddynamicscompat_version` and restores the dependency.

Only that one entry is commented out: the other shadowed dependencies (re2j, netty) are unaffected, and the `slim` artifact the compat needs is unchanged. Verified on a build with no `secrets.properties` overrides, which is what CI does: `build` passes, 1001 game tests pass, the jar contains 0 `integrateddynamicscompat` classes and the 447 re2j/netty classes it should.

The interim 26.3 release therefore ships without the compat integrations until the follow-up lands.

## Tooling

- NeoForge `26.2.0.6-beta` → `26.3.0.7-beta`, NeoForge ModDev `2.0.141` → `2.0.147`.

## Input

`org.lwjgl.glfw` is gone. Key codes now come from `InputConstants`, which carries SDL scancodes rather than GLFW key codes, and `InputConstants.Type.KEYSYM` became `Type.KEYBOARD`.

One of these was not a mechanical swap: the Display Panel's copy shortcut compared a typed **codepoint** against `GLFW_KEY_C`, which happened to equal `'C'` (67). `InputConstants.KEY_C` is an SDL scancode (6) and would never match a codepoint, so that one compares against `'C'` directly.

## Fuel

`FuelValues` was replaced by the `COOKING_FUEL` data component, whose burn time is a `ResolvableInt` resolved against a `CONTAINER_PROCESS` loot context. **Burn times are block-dependent now** — a blast furnace burns the same item faster — so there is no longer a single context-free answer.

`FuelHelpers` builds that context two ways:

- the Coal Generator resolves against itself, which is exact;
- the `burn_time` operator has no block in scope, so it resolves as a plain vanilla furnace would, matching what the operator returned before.

`can_burn` is now a plain `COOKING_FUEL` presence check. Verified in a client: coal 1600, coal block 16000, lava bucket 20000, planks 300, diamond 0 — identical to the old values, and both paths agree.

## Data-driven vanilla systems

- `ComposterBlock.COMPOSTABLES` became the `COMPOSTABLE` data component. The old `0.3` and `0.65` chances map exactly onto `compostable/low` and `compostable/medium` (Menril Berries therefore match vanilla apples). This also drops `neoforge/data_maps/item/compostables.json`, since that data map no longer exists.
- `ItemAbilities.AXE_STRIP` is gone; axe stripping is a `BlockTransformer` now. The two `BlockToolModificationEvent` handlers are replaced by NeoForge's `strippables` data map, which copies shared properties exactly as the handlers did by hand. Verified in a client: a Menril Log with `axis=x` strips to a stripped log still at `axis=x`.

## Worldgen

`Feature`, `ConfiguredFeature` and `FeatureConfiguration` were merged, which touches both code and data:

- `TreeConfiguration` → `TreeFeature`; `configured_feature/` files move to `worldgen/feature/` with their `config` block inlined.
- `TreeGrower` takes `WeightedList`s and a shortest tree type instead of `Optional`s.
- State provider types were renamed: `simple_state_provider` is an inline `{"id": …}`, `weighted_state_provider` → `weighted`, `rule_based_state_provider` → `rule_based`. The below-trunk rule is identical to vanilla's `minecraft:soil_beneath_tree`, so it references that.
- The `random_offset` placement modifier became `offset` with separate `x`/`y`/`z` providers.

Verified by growing a Menril tree from a sapling in a client: 27 logs and 77 leaves, rendering with the usual wide trunk and cyan canopy.

## Blocks and rendering

- The block type `MapCodec` registry and `Block#codec` were removed, so the `CODEC` fields and `codec()` overrides are gone from twelve blocks.
- NeoForge's `canConnectRedstone` was replaced by vanilla's `shouldRedstoneWireConnectTo`, which uses the same direction convention, so the body is unchanged.
- `IClientBlockExtensions#addHitEffects` takes a `BlockPos` and `Direction` instead of a `HitResult`.
- `PoseStack#mulPose(Quaternionf)` → `rotateDegrees(Axis, float)`.
- `BakedQuad.MaterialInfo` gained item glint render types and takes a nullable shade `Direction` instead of a `shade` boolean. `CableModelBase` builds it through `MaterialInfo.of` so the derived item render types stay consistent; `Transparency.NONE` maps to the `ChunkSectionLayer.SOLID` it used before.
- `FaceBakery.bakeQuad` likewise takes a nullable shade `Direction`.
- `PushReaction.DESTROY` → `POPPED`.
- `EnderMan` was renamed `Enderman`, so the access transformer follows.

## Loot and advancements

- `ContextAwarePredicate` was removed; criterion triggers hold `Optional<Holder<LootItemCondition>>` and serialize via `LootItemCondition.CODEC`.
- `LootItemConditionalFunction` takes `Optional<Holder<LootItemCondition>>`.
- `LootContext#getOptionalParameter` became `getOptional`, and `getParameter` is gone entirely, so the wrench condition reads `ORIGIN` optionally and null-checks it, as vanilla's own conditions now do.
- Serialized block values in advancements use the block state NBT key `id` rather than `Name`.

## Other

- `Player#drop` and `Inventory#placeItemBackInInventory` take a `Prediction`. All three call sites are explicitly server-side, so they pass `SERVER_ONLY`, matching vanilla's container code.
- `LivingEntity#randomTeleport` needs an avoidance tag. Liquid Chorus mirrors chorus fruit, so it uses `CONSUMABLE_DOES_NOT_TELEPORT_TO`.
- An apple now carries 17 data components rather than 15: `swing_animation` split into `attack_animation` and `interact_animation`, and `compostable` and `use_effects` are new. `testItemStackDataKeys` now asserts all 17 by name rather than the previous 13 of 15.

## Verified in a client

Cable models and connections, the Coal Generator burning coal and pushing 1340 FE through a cable network into an Energy Battery, axe stripping, composting components, tree growth, and the Logic Programmer GUI — including typing into the search field and arrow-key navigation, which is what the `InputConstants` migration directly affects. No errors in the client log.

## Not verified

The other part and display renderers are compile-checked only; the cable model is the rendering path that was exercised in a running client.

https://claude.ai/code/session_01V35MFZ7JSgoLe79J4d1uxf

---
_Generated by [Claude Code](https://claude.ai/code/session_01V35MFZ7JSgoLe79J4d1uxf)_

---


rubensworks:
CI is red, and will stay red until IntegratedDynamics-Compat is ported to 26.3.

Both `CI` and `Fuzzing` fail at Gradle configuration time, after 19s, before anything is compiled:

```
> Could not resolve all dependencies for configuration ':shadow'.
   > Could not find org.cyclops.integrateddynamicscompat:integrateddynamicscompat-26.3-neoforge:1.0.0-202.
```

There is no fix to port here: that artifact does not exist at any version for 26.3, and the compat module is on the compile classpath (it is `shadow`ed into the jar), so no build can get past configuration. `cyclopscore-26.3` and `commoncapabilities-26.3` are unpublished for the same reason and will surface next once this one resolves.

The change itself is verified locally against those three built from source: `./gradlew build` passes, all 1001 game tests pass, and a 26.3 dev client exercised the cable models, the Coal Generator's fuel path, axe stripping, tree growth and the Logic Programmer's keyboard handling.

Not re-running: a re-run cannot make a missing artifact appear.

---

rubensworks:
The compat side is now ported: CyclopsMC/IntegratedDynamics-Compat#44.

With that branch published to a local maven, this PR builds clean against a genuine 26.3 compat jar (`integrateddynamicscompat-26.3-neoforge:1.0.0-DEV`), so the `:shadow` resolution failure above is resolved once that merges and publishes.

Remaining blockers for this PR are then just the usual three, shared with the rest of the 26.3 set: `cyclopscore-26.3`, `commoncapabilities-26.3`, and now `integrateddynamicscompat-26.3`.

---

src/main/java/org/cyclops/integrateddynamics/blockentity/BlockEntityCoalGenerator.java
```diff
@@ -143,8 +143,11 @@ protected int addEnergyFe(int energy, boolean simulate) {
         return EnergyHelpers.fillNeigbours(getLevel(), getBlockPos(), energy, simulate);
     }
 
-    public static int getFuelTime(ItemStack itemStack) {
-        return itemStack.getBurnTime(RecipeType.SMELTING, ServerLifecycleHooks.getCurrentServer().fuelValues());
+    public int getFuelTime(ItemStack itemStack) {
+        // Burn times resolve against the burning block since 26.3, so they are read through this generator
```
rubensworks:
```suggestion
```

src/main/java/org/cyclops/integrateddynamics/client/gui/container/ContainerScreenPartDisplay.java
```diff
@@ -78,7 +78,8 @@ protected void extractLabels(GuiGraphicsExtractor guiGraphics, int mouseX, int m
 
     @Override
     public boolean charTyped(CharacterEvent evt) {
-        if (GLFW.GLFW_KEY_C == evt.codepoint() && KeyModifier.CONTROL.isActive(KeyConflictContext.GUI)) {
+        // A codepoint, not a key code: GLFW_KEY_C happened to equal 'C', SDL scancodes do not
```
rubensworks:
```suggestion
```

src/main/java/org/cyclops/integrateddynamics/core/evaluate/operator/Operators.java
```diff
@@ -1786,10 +1785,8 @@ public IValue evaluate(OperatorBase.SafeVariablesGetter variables) throws Evalua
             .symbol("burn_time").operatorName("burntime").interactName("burnTime")
             .function(OperatorBuilders.FUNCTION_ITEMSTACK_TO_INT.build(itemStack -> {
                 if (!itemStack.isEmpty()) {
-                    int burnTime = itemStack.getBurnTime(null, ServerLifecycleHooks.getCurrentServer().fuelValues());
-                    return EventHooks.getItemBurnTime(itemStack, burnTime == -1
-                            ? itemStack.getBurnTime(RecipeType.SMELTING, ServerLifecycleHooks.getCurrentServer().fuelValues())
-                            : burnTime, null, ServerLifecycleHooks.getCurrentServer().fuelValues());
+                    // Burn times are block-dependent since 26.3, so a plain furnace is assumed here
```
rubensworks:
```suggestion
```

src/main/java/org/cyclops/integrateddynamics/core/helper/FuelHelpers.java
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraft.world.level.storage.loot.providers.number.ints.ResolvableInt;
+import net.minecraft.world.phys.Vec3;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+/**
+ * Fuel burn times became block-dependent in Minecraft 26.3: {@link CookingFuel} holds a
+ * {@link ResolvableInt} that is resolved against a {@link LootContextParamSets#CONTAINER_PROCESS}
+ * context, so a blast furnace burns the same item faster than a regular furnace.
```
rubensworks:
```suggestion
```

src/main/java/org/cyclops/integrateddynamics/loot/conditions/LootConditionMatchWrench.java
```diff
@@ -22,12 +23,14 @@ public class LootConditionMatchWrench implements LootItemCondition {
 
     @Override
     public boolean test(LootContext lootContext) {
-        ItemInstance itemInstance = lootContext.getOptionalParameter(LootContextParams.TOOL);
-        Entity entity = lootContext.getOptionalParameter(LootContextParams.THIS_ENTITY);
-        BlockPos blockPos = BlockPos.containing(lootContext.getParameter(LootContextParams.ORIGIN));
-        return itemInstance instanceof ItemStack itemStack
+        ItemInstance itemInstance = lootContext.getOptional(LootContextParams.TOOL);
+        Entity entity = lootContext.getOptional(LootContextParams.THIS_ENTITY);
+        // ORIGIN has no required getter anymore, so it is read optionally
```
rubensworks:
```suggestion
```

Issue #613, opened 2018-07-24
Add toggle tutorial


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____

#### Short description:

We should add a tutorial on how state can be saved using the delayer.

PR #1744, opened 2026-09-20
Add a tutorial on saving state with the Delayer

rubensworks:
Closes #613

Adds a "Saving State" tutorial to *On the Dynamics of Integration* that teaches how the Delayer can be used to store state, placed after "The Value of Operators".

## Tutorial content

1. **Variable Delaying**: craft a Delayer, feed it a redstone reader card, and write its history card.
2. **State Latching**: a lamp that is switched on by one button and off by another. The `state` card (`on || (previous && !off)`, with `previous = get_or_default(history, 0, false)`) is placed in the Delayer itself, so the Delayer's history feeds back into the value it stores. `get_or_default` keeps it working while the history is still empty.
3. **State Toggling**: a single-button toggle. A button press lasts several ticks, so a second Delayer (capacity 2) holds the press history and `presses == [false, true]` detects the rising edge for exactly one tick, independent of which Delayer ticks first. `state = choice(edge, !previous, previous)`.

Each step has an advancement (`saving_state/variable_delaying`, `state_latching`, `state_toggling`) with rewards in the book. The two writer-based advancements match the operator tree structurally, so equivalent builds still count, but the latch card does not satisfy the toggle advancement and vice versa.

## Tests

- `GameTestsDelayer`: `testDelayerStateLatch` and `testDelayerStateToggle` build the tutorial network for real (readers, delayers, variable store, redstone writer, lamp), simulate button presses with block changes and assert the lamp over time. This also covers the Delayer feedback loop itself, which had no test before.
- `GameTestsAdvancements`: one test per advancement, plus negative assertions that the latch does not complete the toggling advancement and vice versa.

## Verification

Both constructions were also built and driven in the dev client (button presses as the player, lamp observed on/off/on, advancements granted through the real Delayer GUI) before writing the text. `./gradlew build` and `./gradlew runGameTestServer` pass (997 game tests).

https://claude.ai/code/session_01QYua96wBFk9NEknqGfksfm

---
_Generated by [Claude Code](https://claude.ai/code/session_01QYua96wBFk9NEknqGfksfm)_


src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1764,6 +1770,14 @@
...
 
+    "info_book.integrateddynamics.tutorials.saving_state": "Saving State",
+    "info_book.integrateddynamics.tutorials.saving_state.text1": "Everything you have built so far is &ostateless&r: a variable is always recomputed from the current values of its inputs, so a network has no memory of what happened before. Sometimes you need exactly that memory, for example a lamp that stays on after a button was pressed. In this tutorial, we will use the &lDelayer&r to save state.",
+    "info_book.integrateddynamics.tutorials.saving_state.text2": "A &lDelayer&r remembers the previous values of the &lVariable Card&r in its top slot, and exposes them as a &7List&0. Start by crafting one, attach it to a network with a &lRedstone Reader&r that is targeting a &lButton&r, and insert the reader's &9Boolean&0 &lRedstone&r &lVariable Card&r into the top slot. Set the &lCapacity&r to &61&0 in its gui, and write a new &lVariable Card&r using the bottom slots. This card refers to the &lDelayer&r's history.",
```
rubensworks: Explain why 61 must be selected.

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1764,6 +1770,14 @@
...
+    "info_book.integrateddynamics.tutorials.saving_state.text3": "The trick for saving state is a &ofeedback loop&r: put a &lVariable Card&r in the &lDelayer&r whose value depends on the &lDelayer&r's own history. Each tick, the &lDelayer&r evaluates that card and stores the result, which becomes the previous value on the next tick. Let's use this to build a &olatch&r: a lamp that is switched on by one &lButton&r, and switched off by another.",
+    "info_book.integrateddynamics.tutorials.saving_state.text4": "Read the two &lButtons&r with &lRedstone Readers&r as &9Boolean&0 &lRedstone&r cards &oOn&r and &oOff&r. Next, define &oPrevious&r using the &lGet Or Default&r operator on the history card with index &60&0 and default &9False&0, so that it also works while the history is still empty. Now define &oState&r as &oOn&r &lOr&r (&oPrevious&r &lAnd&r &lNot&r &oOff&r), and place it in the &lDelayer&r's top slot. Place all referenced cards in a &lVariable Store&r. Finally, create a second &oState&r card in the same way, and insert it into a &lRedstone Writer&r that is targeting a &lRedstone Lamp&r. The lamp will now stay on after pressing &oOn&r, until you press &oOff&r.",
```
rubensworks:
```suggestion
    "info_book.integrateddynamics.tutorials.saving_state.text4": "Read the two &lButtons&r with &lRedstone Readers&r as &9Boolean&0 &lRedstone&r cards named &oOn&r and &oOff&r. Next, define a new &lVariable Card&r named &oPrevious&r using the &lGet Or Default&r operator on the history card with index &60&0 and default value &9False&0, so that it also works while the history is still empty. Now define another &lVariable Card&r named &oState&r as &oOn&r &lOr&r (&oPrevious&r &lAnd&r &lNot&r &oOff&r), and place it in the &lDelayer&r's top slot. Place all referenced cards in a &lVariable Store&r. Finally, create a second &lVariable Card&r named &oState&r in the same way, and insert it into a &lRedstone Writer&r that is targeting a &lRedstone Lamp&r. The lamp will now stay on after pressing &oOn&r, until you press &oOff&r.",
```

Besides the suggestion above, also clarify what "in the same way" means above.

src/main/resources/assets/integrateddynamics/lang/en_us.json
```diff
@@ -1764,6 +1770,14 @@
...
+    "info_book.integrateddynamics.tutorials.saving_state.text5": "For the final part, we will make a &otoggle&r that flips on every press of a &osingle&r &lButton&r. If we would simply invert &oPrevious&r while the &lButton&r is pressed, the state would flip on every tick of the press. We therefore need to detect the exact tick in which the &lButton&r &obecomes&r pressed. Place a second &lDelayer&r with &lCapacity&r &62&0, insert a &lRedstone&r card of the &lButton&r, and write its history card. This history is exactly &7[False, True]&0 in the tick after the &lButton&r was pressed.",
+    "info_book.integrateddynamics.tutorials.saving_state.text6": "Define &oEdge&r as this history card &lEquals&r a static &7List&0 of the &9Booleans&0 &9False&0 and &9True&0. Now define &oState&r using the &lChoice&r operator: if &oEdge&r, then &lNot&r &oPrevious&r, otherwise &oPrevious&r. Place it in the top slot of the first &lDelayer&r, and insert a second copy into the &lRedstone Writer&r. Every press of the &lButton&r now toggles the lamp. Feedback loops like these can save any kind of value, not just &9Booleans&0, so you can now build counters, timers and much more.",
```
rubensworks:
```suggestion
    "info_book.integrateddynamics.tutorials.saving_state.text6": "Define a new &lVariable card&r named &oEdge&r as this history card &lEquals&r a static &7List&0 of the &9Booleans&0 &9False&0 and &9True&0. Now define another &lVariable card&r named &oState&r using the &lChoice&r operator: if &oEdge&r, then &lNot&r &oPrevious&r, otherwise &oPrevious&r. Place it in the top slot of the first &lDelayer&r, and insert a second copy into the &lRedstone Writer&r. Every press of the &lButton&r now toggles the lamp. Feedback loops like these can save any kind of value, not just &9Booleans&0, so you can now build counters, timers and much more.",
```

src/main/resources/assets/integrateddynamics/lang/en_us.json
rubensworks: Added the reason in 84818ba: the newest value is always the *last* element of the list, so a Capacity of 1 makes the history hold exactly one element, the value of the previous tick, which is what index 0 then returns. With a larger Capacity, index 0 would be an older value instead.

Checked against `BlockEntityDelay.Ticker`: the queue is FIFO, and `list` is only reassigned *after* the top slot variable is evaluated, so during evaluation the history still holds the previous tick's value.

---

src/main/resources/assets/integrateddynamics/lang/en_us.json
rubensworks: Applied in 84818ba, and replaced "in the same way" with what it actually means: "create a second Variable Card named State by building that same Or expression once more in the Logic Programmer, as the first one has to stay in the Delayer".

---

src/main/resources/assets/integrateddynamics/lang/en_us.json
rubensworks: Applied in 84818ba, with one deviation: the suggestion writes "Variable card", and I used "Variable Card" to stay consistent with the rest of the book. Let me know if you prefer it verbatim.

---

rubensworks:
Play-tested the tutorial end to end in a dev client, building both constructions through the real GUIs (Logic Programmer, reader, writer, Delayer, Variable Store) rather than by script. Both work as promised: the latch stays lit after the On button pops back out and clears on Off, and the single-button toggle flips exactly once per press. All three advancements fired from normal play.

The text, however, turned out not to be followable as written. Fixed in b314f38:

* **One card per operator.** "define State as On Or (Previous And Not Off)" reads like one card, but the Logic Programmer writes one card per operator application. The latch is three cards and the toggle state is two, and the unnamed intermediates also have to go in the Variable Store. Every card is now named (`NotOff`, `Keep`, `State`, `NotPrevious`).
* **Get Or Default takes cards, not inline values.** Its inputs are List, Integer and Any, so "index 0 and default False" means creating an Integer card and a Boolean card first. Nothing said so; the appendix page in the book shows the three inputs, which now matches the text.
* **Capacity 1 is load-bearing and was only mentioned in the warm-up.** With the first Delayer left at the default Capacity 5, the latch still works but the toggle silently stops flipping, with no error and a green check everywhere. Restated where the toggle is built.
* **Which aspect row.** The Redstone Reader offers Redstone Low / Redstone / Redstone High; picking "Redstone Low" gives an inverted latch that looks plausible at first. The text now names the plain Redstone aspect and what it means.
* **Smaller gaps:** the Delayer's top slot already holds the reader card at that point (it has to be swapped), the feedback reference must be indirect (putting the history card in its own Delayer gives "The variable with id N is referring to itself"), the toggle needs a freshly written Redstone card since the On card stays in the store, the static list needs False first and True second, Ticks/Operation must stay at 1, and missing cards in the store surface as "Variable with id N could not be found within the current network".
* Added the **And** and **Not** operator appendices, which the tutorial used but did not show.

The section is now 8 paragraphs instead of 6. Book rendering verified in-game (19 pages, all six operator appendices render). `./gradlew build` and `./gradlew runGameTestServer` pass (997 tests).

---

